PR #1029, opened 2026-05-04
Add Wendland C2 kernel option to density_estimation.py (Issue #1028)

derrring:
Fixes #1028.

## Summary

Adds a `kernel="wendland_c2"` option to `gaussian_kde_gpu` (and the GPU-internal sibling `gaussian_kde_gpu_internal`). Gaussian remains the default; the new option enables kernel-matched FP/HJB pipelines where GFDM stencil construction already uses Wendland $C^2$ weights.

## Kernel form

For support radius $\delta$:

$$K(r) = \frac{3}{2\delta} \left(1 - \frac{|r|}{\delta}\right)_+^4 \left(4 \frac{|r|}{\delta} + 1\right)$$

with $Z_1(\delta) = 2\delta / 3$ (analytic, verified by test).

## Bandwidth semantics

The two kernels interpret the `bandwidth` argument differently:

- `kernel="gaussian"` (default, unchanged): `bandwidth` is factor * std(particles), scipy.stats.gaussian_kde compatible
- `kernel="wendland_c2"`: `bandwidth` is the support radius $\delta$, used directly (no std multiplication)

## Real benefits when matched ships

- **Compact support** $[-\delta, \delta]$ eliminates the Gaussian-tail spillover handled today by the $\mathcal{T}[\mu]$ measure-restriction operator
- **Single-kernel narrative**: one Wendland $C^2$ kernel governs both differential operator approximation (GFDM) and density reconstruction (KDE)

## What this is NOT

- Not "spectral alignment" in any Fourier sense (Wendland has algebraic decay, not Gaussian-exp)
- Not adjoint-consistent with the HJB algebra (no $A_{FP} = A_{HJB}^T$ for hybrid pipelines)

## Test plan

- [x] `pytest tests/unit/test_alg/test_density_estimation.py -v` — 14 passed (7 new + 7 existing). Tests cover: invalid-kernel rejection, $Z_1$ normalization, unit-mass single particle, compact support property, non-negative + finite density, total-mass preservation, Gaussian default unchanged.
- [x] `ruff format` + `ruff check` — clean
- [ ] **mfg-research validation**: matched-vs-mismatched comparison on 1D Boltzmann-Gibbs MFG, sweep $h_{KDE} / \delta_{GFDM} \in \{0.5, 0.75, 1.0, 1.25, 1.5, 2.0\}$. Production decision (paper ships Wendland or Gaussian) follows from the data.

## Out of scope

- Adaptive bandwidth selection for Wendland (Wendland's $\delta$ is naturally the support radius — adaptive selection is a separate research question)
- nD Wendland (`gaussian_kde_gpu_nd`); 1D suffices for the current paper. nD form is straightforward but the dimension-dependent $Z_d$ normalization warrants a separate PR.

### DIFF
--- a/mfgarchon/alg/numerical/density_estimation.py
+++ b/mfgarchon/alg/numerical/density_estimation.py
@@ -39,13 +39,18 @@ def _wendland_c2_normalization_1d(delta: float) -> float:
     inner integral expansion: (1-u)^4 (4u+1) = 1 - 10 u^2 + 20 u^3 - 15 u^4 + 4 u^5,
     so int_0^1 = 1 - 10/3 + 5 - 3 + 4/6 = 1/3.
 
-    Note on naming: this polynomial is Wendland's phi_{3,1}, positive definite
-    in R^d for d <= 3. mfgarchon's GFDM uses the same phi_{3,1} dimension-
-    independently (see WendlandKernel in mfgarchon.utils.numerical.kernels).
+    Polynomial choice: this is Wendland's phi_{3,1}, positive definite in R^d
+    for d <= 3. mfgarchon uses phi_{3,1} dimension-independently throughout:
+    the GFDM WendlandKernel applies the same polynomial in 1D, 2D, and 3D
+    (see mfgarchon.utils.numerical.kernels:426). This KDE follows the same
+    convention deliberately - one polynomial across all dimensions, only the
+    normalization constant Z_d changes per dimension. Z_1 = 2 delta / 3 here;
+    a future nD Wendland KDE would compute Z_d analogously.
+
     The canonical 1D-minimal-degree Wendland C^2 is the different polynomial
-    phi_{1,1} = (1-q)^3 (3q+1), with normalization alpha_1 = 5/(4h). That form
-    is not used here because the goal is kernel-matching with the GFDM
-    weight function, not minimal-degree in 1D.
+    phi_{1,1} = (1-q)^3 (3q+1), with alpha_1 = 5/(4h). That form is NOT used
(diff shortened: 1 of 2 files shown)


derrring:
## Wendland C^2 polynomial choice — clarification (commit 3ced24fc)

@reviewer-self surfaced that "Wendland C^2" is ambiguous: it can refer to either of two distinct polynomials, both with C^2 continuity:

| | Polynomial | Positive definite in | Used in |
|---|---|---|---|
| $\phi_{1,1}$ (canonical 1D-minimal) | $(1-q)^3 (3q+1)$, $\alpha_1 = 5/(4h)$ | $\mathbb{R}^1$ | SPH literature 1D, Dehnen-Aly 2012 |
| $\phi_{3,1}$ (used here) | $(1-q)^4 (4q+1)$, $Z_1(\delta) = 2\delta/3$ | $\mathbb{R}^d, d \leq 3$ | mfgarchon's GFDM `WendlandKernel(k=1)` (used dimension-independently) |

This PR implements $\phi_{3,1}$. The choice is deliberate: the goal is kernel-matching with the GFDM stencil weights, which use $\phi_{3,1}$ in mfgarchon's `kernels.py:426` regardless of dimension. Using $\phi_{1,1}$ here would NOT match GFDM in 1D.

Both forms are mathematically Wendland C^2; the polynomials are different and their normalization constants differ. The docstring (commit 3ced24fc) now flags this explicitly, and the canonical 1D-minimal $\phi_{1,1}$ is noted as a possible follow-up option (`kernel="wendland_c2_d1"`) if a use case for it arises.

Existing 7 unit tests still pass; no math/code change in this commit, just docs.

Issue #1939, opened 2026-08-14
Boundary-name aliases resolve on the ghost path and not on the query path: an absorbing wall silently becomes no-flux


derrring:
`_find_segment_for_face` (the FDM ghost path) resolves a segment's `boundary` string through `types.parse_boundary_face`, which accepts aliases. `get_bc_type_at_boundary` / `get_bc_value_at_boundary` / `matches_point` compare the string with raw `==` / `!=` (`conditions.py:428, 457`, `types.py:602`). So the same `BoundaryConditions` object answers two different questions depending on which route asks.

`_BOUNDARY_STRING_TO_FACE` advertises `left/right/top/bottom`, and `BCSegment`'s own docstring names `"left"` and `"right"` first.

## Measured end-to-end, live 1-D FP-FDM solve

Absorbing exit on the low wall, no-flux elsewhere, `N=21`, `sigma=0.4`, `T=0.2`:

```
canonical  boundary="x_min"/"x_max"
  get_bc_type_at_boundary('x_min') = dirichlet    ('x_max') = no_flux
  final m[0] = 0.000000e+00   m[-1] = 3.900460e-02   mass = 0.245622
alias      boundary="left"/"right"
  get_bc_type_at_boundary('x_min') = no_flux      ('x_max') = no_flux
  final m[0] = 3.900460e-02   m[-1] = 3.900460e-02   mass = 0.248713
  max|difference| = 3.900460e-02
POSITIVE CONTROL  max|canonical - canonical rerun| = 0.000000e+00
```

The absorbing wall silently becomes no-flux. Alias sweep: `bottom`, `left`, `right`, `top` all differ between the two routes (`dirichlet` vs `no_flux`); `x_min`, `x_max`, `y_min`, `y_max` agree.

**The census reviewer's first probe was degenerate and it says so**: with `default_bc=DIRICHLET`, `get_bc_type_at_boundary`'s fall-through returned the right answer for the wrong reason and the two routes agreed. The separating input requires `default_bc` to differ from the segment's type.

## Two adjacent defects in the same owner

**A. `parse_boundary_face` fails silently on unknown prefixes.** Its last branch (`types.py:356-358`) maps *any* string ending `_min`/`_max` to **axis 0** rather than returning `None`. So `"inlet_min"` becomes `BoundaryFace(0, "min")`.

[...]

Issue #577, opened 2026-01-15
Consolidate Neumann BC ghost cell handling into unified infrastructure


derrring:
## Problem

There are **6+ places** in `applicator_fdm.py` handling Neumann BC ghost cells:
- Line 232: `apply_boundary_conditions_2d` 
- Line 1099: `apply_boundary_conditions_3d`
- Line 1255: `apply_boundary_conditions_nd`
- Line 1647: `get_ghost_values_nd`
- Line 1696: Another ghost value path
- Line 2664: `PreallocatedGhostBuffer._update_ghosts_uniform`

Plus `pad_with_ghost_cells()` in `hjb_sl_interpolation.py`.

This duplication leads to:
1. Inconsistent implementations (some had O(h^1) bugs, some correct)
2. Maintenance burden
3. Confusion about which path is used

## Solution

1. **Deprecate** function-based ghost cell APIs
2. **Add concrete replacement** `pad_array_with_ghosts()` function
3. **Consolidate** to `PreallocatedGhostBuffer` as the single source of truth
4. **Update all solvers** to use unified interface

### Phase 1: Deprecation & New API ✅ COMPLETED (2026-01-15)
- [x] Fix reflection formula in `_update_ghosts_uniform`
- [x] Add deprecation warnings to function-based APIs (v0.19.0 removal)
- [x] Add `pad_array_with_ghosts()` concrete replacement function
- [x] Export from `mfg_pde.geometry.boundary`
- [x] Update deprecation roadmap documentation

### Phase 2: Update Solvers ✅ COMPLETED (2026-01-16)
- [x] Verify Semi-Lagrangian uses ghost buffer (not fallback) - Uses characteristic tracing, no changes needed
- [x] Update WENO to use unified interface - Completed in #576 (uses `PreallocatedGhostBuffer` with order=5)
- [x] Update FDM solvers to use unified interface - Migrated to `pad_array_with_ghosts()`
  - [x] Tensor Calculus (`tensor_calculus.py`)
  - [x] HJB FDM Solver (`hjb_fdm.py`)

### Phase 3: Cleanup (v0.19.0)
- [ ] Remove deprecated function-based code
- [ ] Update documentation

## Current Status

**Phases 1-2 Complete** (2026-01-16). All solvers now use unified BC interface:
- ✅ FDM: Uses `pad_array_with_ghosts()`
- ✅ WENO: Uses `PreallocatedGhostBuffer` with high-order support (#576)
[...]

PR #1955, opened 2026-08-15
refactor(geometry): _compat keeps only its ghost-value entry point

derrring:
Stage 0a of the BC dispatch consolidation. **19 → 13** `BCType` dispatch chains, −2213/+76 lines.

## What this is

A census of `mfgarchon/geometry/boundary/*.py` finds **19 distinct `if/elif` chains dispatching on `BCType`**, over an enum with 8 members. No two agree:

| | count |
|---|---|
| chains covering all 8 members | **2** / 19 |
| chains mentioning `EXTRAPOLATION_*` | 4 / 19 |
| chains missing `ROBIN` | 7 |
| chains missing `REFLECTING` | 5 |
| chains ending in a silent `else` or **no `else` at all** | **19** / 19 |
| distinct coverage signatures | 8 over 19 chains |
| chains in `_compat.py` | **8** / 19 |

Not one of them raises on an unrecognised type at the chain level, so an unhandled BC type produces a wrong answer rather than an error. This PR takes the largest block that can go **without a design decision behind it**: six of the eight chains inside the already-deprecated `_compat.py`.

## What went, and what did not

**Deleted** — `apply_boundary_conditions_1d/2d/3d/nd`, `create_boundary_mask_2d`, twelve private helpers. 1453 of the module's 1856 lines. Their public export was already withdrawn (#577 Phase 3), no library module imports them, and the only importers left were two test files reaching into the private module by name.

**Kept** — `get_ghost_values_nd` and its five-function closure. It is the one name still reachable as an attribute of `mfgarchon.geometry.boundary`, via that module's `__getattr__`, and its removal is **declared for v0.25.0 while the tree is at `0.22.0.dev0`**. Deleting it now would break a published contract three minor versions early.

Its output is pinned **byte-identical** across this change over five cases — uniform Dirichlet / Neumann / periodic, a mixed BC, and 1D — captured before the edit and re-checked after (`sha256:62a79ecb…`).

## A grep by name overcounts this tenfold

Counting bare occurrences of the five public names gives **82** sites. Resolving imports gives **9**, all in two test files. The gap is a name collision: `hjb_sl_characteristics.py:174,289` defines its own `apply_boundary_conditions_1d` / `_nd`, which clamp characteristic foot points back into the domain and share nothing with the field-padding functions of the same name here. The semi-Lagrangian solvers import that pair, not this one.

## Coverage this removes and does not replace

No test now asserts that `BCType.EXTRAPOLATION_LINEAR` or `EXTRAPOLATION_QUADRATIC` **routes to** its ghost formula. The formulas stay pinned directly (`ghost_cell_linear_extrapolation` / `_quadratic` have their own tests), but the only surviving dispatch that reaches them is `bc_to_topology_calculator`, which has no test at all. Stated here rather than papered over.

## Docs

The ghost-cell examples move to `pad_array_with_ghosts`. Two "Common Issues" entries are **deleted rather than translated** — both documented errors of the removed `apply_boundary_conditions_2d`, and neither reproduces on the canonical path. Measured: an unbound BC and a mixed BC without `domain_bounds` both pad without raising, element-wise identically to the bound and bounded forms, because the dimension comes off the array and face segments are matched symbolically through `parse_boundary_face` (#1939).

## Ratchets

Both moved in the good direction and were re-baselined in the same commit: `silent_pass` 88 → 85, doc `bad_imports` 78 → 77.

## Gate

`./scripts/local_ci.sh` → **6061 passed, 12 skipped, 21 xfailed, GATE GREEN**.


derrring:
**Correction to the PR body.** The table row *"chains ending in a silent `else` or no `else` at all — **19** / 19"* is wrong. Computed rather than read off the matrix: **14 of 19**. Five raise — `_compat.py`'s `_apply_uniform_bc_2d`, `apply_boundary_conditions_nd` and `_apply_bc_1d`, and both of `applicator_meshfree`'s chains.

Three of those five are inside `_compat.py`, so the six chains this PR deletes are **not** a block of uniformly silent ones. The changelog fragment is corrected in `7c2fd4e`; the sentence "Not one of them raises on an unrecognised type at the chain level" above should be struck.

Two further figures the PR body did not state, because the census was scoped to `geometry/boundary/*.py` and that limit was never declared:

| over all of `mfgarchon/` | |
|---|---|
| `BCType` dispatch chains | **26** (19 here, 7 in `alg/` and `operators/`) |
| ending silently | 18 / 26 |
| functions comparing ≥3 BC **string** literals — a second vocabulary the enum census cannot see | **35** |

None of this changes what the PR does or the byte-identity pin on `get_ghost_values_nd`.

Issue #1907, opened 2026-08-12
Robin's low wall applies the outward normal twice: robin_bc(g, alpha=0, beta=1) is not neumann_bc(g) there


derrring:
`BCType.ROBIN` and `BCType.NEUMANN` express the same condition when `alpha = 0, beta = 1`, and they disagree at the LOW wall by exactly `2*h*g`. The high wall agrees to machine precision.

Found by independent review of #1906; **pre-existing**, not introduced by it. Two mutually unaware reviewers reached it by different oracles (analytic continuation of a linear field; the closed form of `u(x) = x^2 - 2x`), and I reproduced it directly before filing.

## Measured

`u = 0.5*cos(2*pi*x)` on `[0,1]`, `g = 2.0`, spacing threaded:

| Nx | `neumann` ghost_lo | `robin(0,1)` ghost_lo | difference | `2*h*g` | ghost_hi difference |
|--:|--:|--:|--:|--:|--:|
| 21 | 0.600000 | 0.400000 | 0.200000 | 0.200000 | 0.0 |
| 81 | 0.525000 | 0.475000 | 0.050000 | 0.050000 | 0.0 |
| 321 | 0.506250 | 0.493750 | 0.012500 | 0.012500 | 0.0 |

The disagreement is exactly `2*h*g`, confined to the low wall, and invariant under the spacing fix (on base `4c5c1840` with the `dx = 1.0` fallback it is `4.0 = 2*1.0*g`).

## Mechanism

The outward normal at the low wall points in `-x`, so `du/dn = -du/dx`. With the ghost at `x_g < x_i`, `du/dx = (u_i - u_g)/dx`, hence

    du/dn = (u_g - u_i)/dx        <- already carries the sign; no further factor is due

`applicator_fdm.py:1174` states the convention as `du/dn = (u_g - u_i)/dx * outward_sign` and applies `outward_sign = -1` at the low wall, which yields `du/dx`. The coefficients follow from it:

    applicator_fdm.py:1182    coeff_ghost_lo = alpha / 2.0 - beta / dx

The NEUMANN branch does not have this problem:

    applicator_fdm.py:1153    buf[tuple(lo_ghost)] += dx * v   # Issue #1262: was -= (du/dx sign), now += (du/dn sign)

So **#1262 fixed exactly this sign on the NEUMANN branch and did not fix it on ROBIN.** `tests/unit/test_bc_applicator.py:1285-1288` records that fix.

## Consequence

[...]

Issue #1456, opened 2026-06-26
[Design/RFC] BC-capability audit: BC-blind shortcuts blocking universal boundary-condition support


derrring:
## Context
Roadmap: make **all BCs available to all solvers**. Today many solvers are implicitly BC-restricted (only ever run no-flux/Neumann), and several "safe-because-no-flux-only" shortcuts become **silent wrong-answers** the moment a non-conserving / non-Neumann BC reaches them. A 12-solver investigate→verify audit (adversarially verified; 3 solvers partially via agent failure, filled by hand) maps the BC-blind shortcuts and the resolution per item.

## The bug class (three kinds)
- **(a) Hardcoded boundary stencils** — zero-flux/Neumann diffusion rows applied regardless of the problem BC (so Dirichlet/Robin/absorbing get the wrong wall operator).
- **(b) Unconditional mass renormalization** — re-pinning density to constant/previous mass every step. Correct only for mass-conserving BC (no-flux/periodic); under absorbing/Dirichlet/Robin-with-flux it **manufactures false conservation and hides the absorption**.
- **(c) Silent BC fallbacks** — resolution cascades that default to `no_flux_bc`/`neumann_bc` (or `None`→clamp) when a BC doesn't resolve, and shared `bc_utils` collapses (robin→reflect, mixed→segments[0], non-periodic→neumann) that silently drop flux/value.

## Per-solver verdict (universal-BC risk)
| Solver | Risk | BC resolution | Headline BC-blind shortcuts |
|---|---|---|---|
| **HJBSemiLagrangian** | **high** | inherited single source ✓ | `_get_diffusion_bc_type` collapses every non-periodic BC→neumann on default `adi`; the periodic geometric fold is **dead** (`bc_type_to_geometric_operation` returns `'periodic'`, code tests `'wrap'`) → periodic feet silently extrapolated; `applicator_*` mixed/robin paths drop flux |
| **FPSLSolver** (forward/adjoint) | **high** | local resolver (2/5 sources) | diffusion BC→neumann collapse; zero-flux CN/ADI stencils; **clamp-advection + splat-clamp = structural re-pin to constant mass** (class b without an explicit renorm line); robin→reflect; resolver can diverge from HJB |
| **FPSLJacobian** (deprecated) | **high** | private 3-cascade | hardcoded Neumann CN stencil + **unconditional mass renorm** (this is S0-11, re-dispositioned) |
[...]

PR #1957, opened 2026-08-16
fix(geometry): Robin's outward normal (#1907), and the FP resolver resolves instead of renaming

derrring:
Two commits. The first is a live defect that blocked the second.

---

## 1 — `14361fe6` Robin's `du/dn` is the outward normal at both walls (#1907)

**The shipped FDM path violated the Robin condition the package declares, at the low wall only.** Measured on `main` through `pad_array_with_ghosts`, `alpha=1, beta=0.3, g=0.7`:

| wall | residual of `alpha*u_b + beta*du/dn - g` |
|---|---|
| min | **6.17** |
| max | −6.7e−16 |

The low wall applied the outward sign a second time, imposing `alpha*u + beta*du/dx = g` — the **axis** condition, a physically different wall. On a cell-centred grid the ghost lies outside at *both* walls, so `(u_g - u_i)/dx` already **is** the outward normal derivative.

**Four implementations, two conventions.** `ghost_cell_robin` and `BaseStructuredApplicator._compute_ghost_robin` were byte-for-byte the same correct side-free form; `PreallocatedGhostBuffer._apply_linear_reflection` and `._apply_ghost_for_face` each carried the extra sign. `ghost_cell_robin` is now the owner — implementations **4 → 2**, and the survivor in `_compat.py` is deleted by #1955.

**User-visible:** `uniform_bc(ROBIN, alpha, beta, value)` reads as "the same condition on every wall" and was not, whenever `beta != 0`. `robin_bc(g, alpha=0, beta=1)` and `neumann_bc(g)` are one condition and disagreed at the low wall by exactly `2*dx*g`.

**What did not change:** 9 of 13 probed cells byte-identical — every other `BCType`, and `ROBIN` with `beta = 0` where the derivative term vanishes. Those are pinned as unchanged.

**Second behaviour change, from the deletion:** two removed implementations silently mirrored when `alpha/2 + beta/dx == 0`. The owner raises — a singular coefficient means the condition doesn't determine the ghost.

**The pin is the condition, not a path comparison** — after routing every site through one owner, path-A-vs-path-B is tautological and passes over a broken owner. Mutations measured: axis convention restored → **15 of 20 failed**; silent mirror restored → **1 failed**.

Two existing tests are **re-pointed, not deleted**. `test_ghost_spacing_1904` predicted this in its own docstring: *"it will need re-pointing when #1907 lands."* `test_fp_particle_gradient_bc_1255` pinned **both** conventions, one per wall, because it recorded what the code did rather than what the condition says.

---

## 2 — `1deb7532` The FP resolver resolves instead of renaming

**Layer 2 has existed since #856 and is unreachable.** Traced, not inferred:

| symbol | library callers |
|---|---|
| `HJBResolver` / `FPResolver` / `BCResolver` in `alg/` | **0** |
| `resolved_bc_to_calculator` | **0** (module demo + 1 test) |
| `resolved_to_boundary_conditions` | **0** |

**Why it stalled is in its own source:**

```python
ZERO_FLUX = "zero_flux"   # J*n = 0, needs drift+diffusion for calculator
```

`FPResolver` "resolved" the impermeable wall by **renaming the intent**, so Layer 3 dispatched on it again. It could not do better: `ResolvedBC.alpha/beta` were `float`, and the Robin coefficient of a reflecting FP wall is `D_pH(x, ∇u)·n` — a **field**.

**The fix.** `ResolvedBC` coefficients accept fields; `FPResolver` emits `ROBIN` with `alpha = v_n`, `beta = -D`, `g = 0`. Verified against the condition:

```
16 scalar coefficient sets      worst |J.n| = 3.6e-15
field-valued drift over a wall  worst |J.n| = 2.2e-16, each point uses its own coefficient
scalar path                     bit-identical
```

`MathBCType.ZERO_FLUX` is **deleted** — the enum's docstring says it "contains no physical intent types" and this was the exception to its own rule. `ZeroFluxCalculator` stays, still reached via `bc_to_topology_calculator(use_zero_flux=True)`.

**No default for a mathematical parameter.** Resolving an impermeable wall without `drift`/`diffusion` raises, naming the missing key. A defaulted drift makes the wall pure Neumann — conserves mass at a tangential wall, leaks at every other one, and **the wrong answer still converges**.

---

## Context

This is step 1 of the plan in the Joplin note *"BC classification for MFG — settled from sources"* §10. The classification was settled from the literature this session: one physical wall imposes **two structurally different conditions**, `∂_n u = 0` on HJB and a **Robin** condition on FP whose coefficient depends on the solution (Carmona–Delarue I §4.7 (4.169)/(4.170); Cirant *JMPA* 103 (2015) system (MFG) *hjn*)/*kn*)). The field calls both "Neumann" — harmless in prose, fatal once a name drives a dispatch.

**Not in scope:** routing `alg/` callers through the resolver and deleting the 26 chains (step 2), and the `σ = 0` regime gate (step 3).

## Gate

`./scripts/local_ci.sh` → **6110 passed, 12 skipped, 21 xfailed, GATE GREEN**.
Also run: `check_internal_deprecation`, `check_docs_structure`, `check_critical_issues` — the CI-only checks the local gate does not cover.

### DIFF
--- a/changelog.d/1907-robin-outward-normal.fixed.md
+++ b/changelog.d/1907-robin-outward-normal.fixed.md
@@ -1,0 +1,37 @@
+**Robin's `du/dn` is the outward normal derivative at both walls.** The low wall applied the
+outward sign a second time, imposing `alpha*u + beta*du/dx = g` — the axis condition, a
+physically different wall. Measured on `main`, `alpha=1, beta=0.3, g=0.7`: the residual of the
+condition `BCSegment.beta` declares was **6.17 at the min wall** and 0 at the max.
+
+On a cell-centred grid the ghost lies outside at *both* walls, so the quotient toward it,
+`(u_g - u_i)/dx`, already is the outward normal derivative; no further factor is due.
+`ghost_cell_robin` owns that derivation and the three sites that restated it now call it:
+`PreallocatedGhostBuffer._apply_linear_reflection`, `._apply_ghost_for_face`, and
+`BaseStructuredApplicator._compute_ghost_robin`.
+
+**User-visible consequence.** `uniform_bc(ROBIN, alpha, beta, value)` reads as "the same
+condition on every wall" and was not: the two walls carried different physics whenever
+`beta != 0`. `robin_bc(g, alpha=0, beta=1)` and `neumann_bc(g)` are the same condition and
+disagreed at the low wall by exactly `2*dx*g`.
+
+**What changed and what did not.** Of thirteen probed cells, **nine are byte-identical** — every
+`BCType` other than `ROBIN`, plus `ROBIN` with `beta = 0`, where the derivative term vanishes and
+the two conventions coincide. The four that moved are all `ROBIN` with `beta != 0` at the low
+wall, and each now satisfies the declared condition to `3e-14` or better across six coefficient
+sets including negative `beta` and negative `g`.
+
+**Second behaviour change, from the deletion.** Two of the removed implementations silently
+mirrored the interior cell when `alpha/2 + beta/dx == 0`. A singular coefficient means the
+condition does not determine the ghost, so the surviving owner raises instead of inventing an
+answer.
+
+The pin is the **condition itself**, not a comparison between paths: once all sites route through
+one owner, path-A-vs-path-B is tautological and would pass over a broken owner. Two mutations
+were measured to redden it — restoring the axis convention in the owner (15 of 20 failures) and
+restoring the silent mirror (1).
+
+Two existing tests are re-pointed rather than deleted. `test_ghost_spacing_1904.py` had said so
+itself: *"the sign convention this asserts is the applicator's own … which #1907 reports is one
+factor too many there … it will need re-pointing when #1907 lands."* And
+`test_fp_particle_gradient_bc_1255.py` pinned **both** conventions, one per wall, because it
+recorded what the code did rather than what the condition says.
--- a/changelog.d/1957-resolved-bc-field-coefficients.changed.md
+++ b/changelog.d/1957-resolved-bc-field-coefficients.changed.md
@@ -1,0 +1,28 @@
+**`FPResolver` resolves the impermeable wall instead of renaming it.** `NO_FLUX` and
+`REFLECTING` now become `MathBCType.ROBIN` carrying the coefficients of the flux condition:
+`J.n = 0` with `J = v*m - D*grad(m)` is `v_n*m - D*d_n m = 0`, i.e. `alpha = v_n`, `beta = -D`,
+`g = 0`. Verified against the condition itself, not against the previous code path: over 16
+coefficient sets the resulting ghost leaves `|J.n| <= 3.6e-15`.
+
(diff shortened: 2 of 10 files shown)


derrring:
**Corrections to the PR body, from independent review.** Four numbers in it are wrong; the code claims hold.

**1. "axis convention restored → 15 of 20 failed" is not that mutation.** Restoring the axis convention means restoring main's applicator, which differs from the fix *only at the low wall*: `git checkout 2f399d56 -- applicator_fdm.py` → **10 failed, 10 passed**. The 15 came from flipping `coeff_ghost`'s sign unconditionally inside `ghost_cell_robin`, breaking **both** walls — a different, over-broad mutation. Discrimination still holds (10 ≫ 0); the label was wrong.

**2. "implementations 4 → 2, and the survivor in `_compat.py` is deleted by #1955" is false on both halves.** After all three PRs merge the real count is **1**.

**3. The headline residuals are pinned by nothing.** `test_resolution.py` asserts only `alpha == 0.6`, `beta == -0.35`, `value == 0.0` — no test evaluates `J·n` on a ghost built from a resolved BC. The "16 coefficient sets, worst 3.6e-15" and "field drift, worst 2.2e-16" were one-off probes; the review reproduced the field figure exactly and got 1.53e-14 over its own 16 sets. Same order, set-specific number.

**4. Blast radius is much smaller than the body implies.** `BCType.ROBIN` reaches no ghost formula in any shipped solve: it is in `_SUPPORTED_BC_TYPES` for exactly one solver (`HJBGFDMSolver`, which uses `_build_neumann_bc_row`), and every other solver refuses it at construction under #1456. The changed code is reachable only through the direct public helpers. "Changes published numerical behaviour for any user with a mixed Robin BC" is true only for direct API users.

**One behaviour change the body does not name, and should.** Under `create_adjoint_consistent_bc_1d` (Nx=21, σ=0.4) the boundary Laplacian moves by 3×:

| | main | this PR |
|---|---|---|
| `du/dn − g` at min wall | −1.3665 (= −2·g) | 3.2e−15 |
| `_compute_laplacian_1d(U)[0]` | −42.596 | −13.898 |

The provider's `g = −σ²/2·∂ₙ ln m` is already outward-normal, so main applied it with the wrong sign at the low wall and this PR fixes it. Consequence: `AdjointConsistentProvider`'s docstring carries published validation numbers (AC 1.36 vs Neumann 2.09; interior 5.88 vs 1.55) produced under the pre-fix convention, now **stale**.

**Still open, not addressed here** — each is pre-existing or narrow, and I would rather file than widen this PR:
- `ghost_cell_robin`'s **vertex-centred** branch still uses `outward_normal_sign` and is wrong under any reading (residual 5.304 at the min wall). The newly-declared owner holds two conventions.
- `RobinCalculator` is still side-dependent on vertex-centred grids, and returns a boundary value rather than a ghost.
- `particle_action_for_bc_type` takes `alpha`, requires it non-`None`, and never reads it.
- `get_ghost_values_nd` (kept by #1955 until v0.25.0) ships a Robin wall that does not satisfy the Robin condition — residuals 1.100 / 1.500.

Blocker fixed in `2cbed01f`.

Issue #1966, opened 2026-08-16
ghost_depth >= 2: Robin and both extrapolations collapse to a constant, Dirichlet writes its layers reversed [CORRECTED -- see #1967]


derrring:
`PreallocatedGhostBuffer` fills more than one ghost layer for high-order schemes. At `ghost_depth = 1` every `BCType` is correct. At `ghost_depth >= 2`, **five of the eight are not**, in two distinct ways.

Found while tracing the dispatch chains for the resolver consolidation. **Pre-existing.** Adjudicated against exact continuations, not against another code path.

## The oracle

Cell-centred grid, `N = 8` cells on `[0,1]`, `h = 0.125`. Ghost layer `k` (k=1 nearest the wall) has its centre at `x_b - (k - 1/2)h`. For a boundary condition that a given field satisfies **exactly**, the correct ghost is that field evaluated at those centres — no discretisation error, so any deviation is the defect.

Three fields, each exact for the condition it is paired with:

| condition | field | why exact |
|---|---|---|
| `EXTRAPOLATION_LINEAR` | `3x + 1` | zero second derivative |
| `EXTRAPOLATION_QUADRATIC` | `2x² − x + 0.5` | zero third derivative |
| `NEUMANN(0)` | `cos(2πx)` | even about `x = 0`, so `du/dx = 0` there |
| `ROBIN(α=1, β=0.5, g=−0.5)` | `3x + 1` | satisfies `αu + β du/dn = g` at the wall by construction |
| `DIRICHLET(1.0 at x=0, 4.0 at x=1)` | `3x + 1` | the field's own boundary values |

## Defect 1 — Robin and both extrapolations collapse to a constant

Every layer receives the value computed for **layer 1**:

```
EXTRAPOLATION_LINEAR   g=1  ghosts=[0.8125]                  exact=[0.8125]                 err 0
                       g=2  ghosts=[0.8125 0.8125]           exact=[0.4375 0.8125]          err 3.75e-01
                       g=3  ghosts=[0.8125 0.8125 0.8125]    exact=[0.0625 0.4375 0.8125]   err 7.50e-01

EXTRAPOLATION_QUADRATIC g=2  ghosts=[0.5703 0.5703]          exact=[0.7578 0.5703]          err 1.88e-01
                        g=3  ghosts=[0.5703 0.5703 0.5703]   exact=[1.0078 0.7578 0.5703]   err 4.38e-01

[...]

PR #1969, opened 2026-08-16
fix(geometry): pair the high wall's ghost layers nearest-first, and scale the flux offset with the layer (#1967)

### LINKED ISSUE #1967: The high wall pairs its ghost layers backwards: NEUMANN/NO_FLUX/REFLECTING are reversed at ghost_depth >= 2, and WENO5 is a live consumer
`PreallocatedGhostBuffer._apply_linear_reflection` walks the low-wall ghost slots **nearest-first** and the high-wall slots **outermost-first**, while pairing both against interiors walked nearest-first. The low wall is correct; the high wall receives its layers in reverse order.

Found by four independent oracles run in parallel on this question (convergence order, invariance, cross-implementation, grid convention), each of which reached it by a different route. Reproduced directly before filing. **Pre-existing.**

## The two loops

`mfgarchon/geometry/boundary/applicator_fdm.py:1154-1172`

```python
for k in range(g):
    lo_ghost[axis]    = g - 1 - k      # g-1, g-2, ... 0   -> nearest ghost first
    lo_interior[axis] = g + k          # g,   g+1,  ...    -> nearest interior first
    buf[lo_ghost] = buf[lo_interior]                        # paired correctly

for k in range(g):
    hi_ghost[axis]    = -(k + 1)       # -1,  -2,   ...    -> nearest ghost first
    hi_interior[axis] = -(g + k + 1)   # -(g+1), -(g+2)... -> FARTHEST interior first
    buf[hi_ghost] = buf[hi_interior]                        # paired backwards
```

At `g = 3` the high wall's `k = 0` gives ghost `-1` (nearest the wall) paired with interior `-4` (three cells in). The correct partner is `-1` ↔ `-1 - g = -4`… no: the mirror of ghost layer 1 is the *first* interior cell, index `-(g+1)` only when `g = 1`. The pairing must be `hi_ghost = -(g - k)` against `hi_interior = -(g + 1 + k)`, or equivalently keep the ghost index and reverse the interior walk.

## Measured

`u = cos(2πx)` is even about **both** `x = 0` and `x = 1`, so `NEUMANN(0)` is exact at both walls and the correct ghost is the field itself at the ghost centres. `N = 8`, `h = 0.125`, cell-centred (the geometry this branch uses — see #1967):

```
g=1  low  err 0.000e+00     high err 0.000e+00
g=2  low  err 0.000e+00     high err 5.412e-01     reversed(got) == want  True
g=3  low  err 0.000e+00     high err 1.307e+00     reversed(got) == want  True
```

`reversed(got) == want` at every depth is what separates a layer-indexing error from wrong arithmetic — the values are all correct, they are stored in the wrong slots.

Independent confirmations, each by a different oracle:

| oracle | signal |
|---|---|
| convergence order | high-wall observed order 2 → **0** at `g ≥ 2`; low wall climbs 2 → 4 → 6 as designed |
| invariance | mirror residual `1.551e+00` / `2.829e+00`; depth-consistency violated at `hi` with `lo` exactly 0 |
| Jacobian | nearest high ghost reads `u[N-g]` instead of `u[N-1]` |
| flux read-back | high wall returns `-80` / `-160` for a requested `du/dn = 0` |

## Why this one matters and the others do not

`hjb_weno.py:330` sets `self.ghost_depth = 3` and reads layers 2 and 3. It is the **only** production consumer of `ghost_depth ≥ 2` in the library, and its `_SUPPORTED_BC_TYPES` is `{NEUMANN, NO_FLUX, PERIODIC}` — precisely the family this defect hits. The other `ghost_depth ≥ 2` defects (#1966) are on members no consumer reaches at that depth.

Note the WENO path takes `_apply_poly_extrapolation` for a *uniform* BC (`order = 5 > 2`), which is clean. This defect is reached when the BC is written **per-face** rather than uniform, or when `order ≤ 2` is selected — measured, the same reversal appears on the per-face path `_apply_ghost_for_face` and in the `GhostBuffer` Topology/Calculator route.

## Correction to #1966

#1966 certifies `NEUMANN`, `NO_FLUX` and `REFLECTING` as "exact at every depth tested (g = 1, 2, 3, machine zero)". That was measured at the **low wall only**, where it is true. The claim is corrected there; this issue is the defect it missed.

Two further corrections recorded in #1966 itself: its count ("five of the eight") and its priority framing ("a silently first-order boundary does the most damage exactly where the high-order schemes need it") are both wrong — with this family included the count is seven of eight, and the high-order path is the clean one.

## Scope

One index expression. The fix owes the oracle above at both walls and every depth, plus a control that the low wall does not move.

Related: #1966 (the latent siblings), #1968 (the class holds two geometries), #1935.

derrring:
Closes #1967.

Two defects in one loop, and the loop has two copies.

## Defect 1 — the high wall pairs its layers backwards

The high ghosts occupy padded indices `-g .. -1`, and **`-g` is the one adjacent to the wall**, so both the ghost walk and the interior walk must start there.

```python
lo_ghost = g - 1 - k        lo_interior = g + k          # nearest ↔ nearest      correct
hi_ghost = -(k + 1)         hi_interior = -(g + k + 1)   # nearest ↔ FARTHEST     reversed
```

At `g = 1` the two expressions coincide. At `g >= 2` the layers arrive in reverse order.

**Measured.** `cos(2πx)` is even about *both* `x = 0` and `x = 1`, so `NEUMANN(0)` is exact at both walls and the correct ghost is the field itself at the ghost centres:

| | low wall | high wall |
|---|---|---|
| `g=1` | 0 | 0 |
| `g=2` | 0 | **5.41e−01** |
| `g=3` | 0 | **1.31e+00** |

`reversed(got) == want` at every depth — **every value correct, every slot wrong**, which is what separates an indexing error from bad arithmetic. After the fix: `4.4e−16` at both walls, all depths.

## Defect 2 — the flux offset does not grow with the layer

Ghost layer `k` sits `(2k−1)·dx` from its mirror, so an inhomogeneous Neumann offset scales with the layer. `dx · v` was the `k = 1` value applied to every layer, drifting by `(2k−2)·dx·v`: **0 at `g=1`, 5.0e−01 at `g=2`, 1.0e+00 at `g=3`** on `f = −2x+1` with `du/dn = 2`.

This one was missed by the first pass, which tested only `du/dn = 0` — where the term vanishes. Non-discriminating input.

## Both copies

`_apply_ghost_for_face` (the per-face path) carried both. Fixing only the uniform copy made the two disagree and reddened `test_per_face_bc_path_1937::test_the_flux_is_applied_at_every_ghost_layer` — **that test doing its job**, and the reason it is not the oracle here: once both copies are corrected, agreement is compatible with both being wrong.

## Why a suite of 6147 was green

**Both defects are invisible at `ghost_depth = 1`** — one layer has no order to reverse, and `(2·1−1)·dx = dx`. Every caller in the library passes 1 or omits it.

The one production consumer at depth 3 is `hjb_weno.py:330`, and its `_SUPPORTED_BC_TYPES` is `{NEUMANN, NO_FLUX, PERIODIC}` — **exactly the family defect 1 hits**.

## Scope, pinned

Eighteen of twenty-four probed cells are unchanged, every `ghost_depth = 1` cell among them, and that is asserted: since every caller uses depth 1, a change there would be a regression in everything that currently works and nothing else in the file would catch it.

The discrimination ratchet's `neumann_low_wall_flux_sign` anchor is re-pointed to the new line — it failed because the source moved, which is what that check exists for.

## Mutations

One per defect per copy:

```
uniform  order reversed again      9 failed / 31 passed
uniform  flux offset constant      4 failed / 36 passed
per-face order reversed again      7 failed / 33 passed
per-face flux offset constant      1 failed / 39 passed
```

Source restored byte-identical after each.

## Provenance

Found by **four independent oracles run in parallel** on #1966 — convergence order, invariance, cross-implementation, grid convention. #1966 had certified this family as correct after measuring **one wall**; its correction is recorded there, along with three others (the count, the priority framing, and a `grep` false negative that hid the live consumer).

One method note: mirror invariance, nominated as the sharpest test of a layer-order defect, is **structurally blind** to a defect that reverses both walls symmetrically. Depth-consistency catches that one — and is itself blind to the Robin collapse, since a layer-independent value is trivially depth-consistent. No single oracle here saw everything.

## Gate

`./scripts/local_ci.sh` → **6187 passed, 12 skipped, 21 xfailed, GATE GREEN**.

### DIFF
--- a/changelog.d/1967-ghost-layer-order-and-flux.fixed.md
+++ b/changelog.d/1967-ghost-layer-order-and-flux.fixed.md
@@ -13,9 +13,17 @@ must scale with the layer. `dx · v` was applied to every layer — the `k = 1` 
 throughout, drifting by `(2k-2)·dx·v`.
 
 **Both are invisible at `ghost_depth = 1`**, where a single layer has no order to reverse and
-`(2·1−1)·dx = dx`. Every caller in the library passes 1 or omits it, which is why a suite of 6147
-was green with both present. The one production consumer at depth 3 is `hjb_weno.py:330`, whose
-`_SUPPORTED_BC_TYPES` is `{NEUMANN, NO_FLUX, PERIODIC}` — exactly the family the first defect hits.
+`(2·1−1)·dx = dx`. `hjb_weno.py:330` is the library's only production `ghost_depth >= 2`, so every
+other caller sits in that blind spot, which is why a suite of 6147 was green with both defects
+present.
+
+**Reachability, measured rather than inferred.** WENO's `_SUPPORTED_BC_TYPES` is
+`{NEUMANN, NO_FLUX, PERIODIC}` — the family these defects hit — but on every *uniform* BC it
+admits, it dispatches to `_apply_poly_extrapolation`, which neither defect touches and this change
+leaves byte-identical. The corrected lines are reached only through a *per-face* BC, and no
+in-repo code hands WENO one: the two per-face HJB constructors both emit `BCType.ROBIN`, which
+WENO refuses at construction. So this is a correctness fix on a path that is reachable but not
+currently reached — not, as an earlier draft of this fragment said, the live consumer's own path.
 
 Eighteen of twenty-four probed cells are unchanged, every `ghost_depth = 1` cell among them, and
 that is pinned: since every caller uses depth 1, a change there would be a regression in
@@ -24,3 +32,7 @@ everything that currently works.
 Found by four independent verification oracles — convergence order, invariance,
 cross-implementation and grid convention — run in parallel on #1966, which had certified this
 family as correct after measuring one wall.
+
+A **third** copy of the same reversal survives in `GhostBuffer._update_bounded`, at both walls and
+with identical magnitudes; it is unreachable today (`create_ghost_buffer_from_bc` has no library
+caller) and is filed separately. This change corrects two of three.
--- a/changelog.d/1967-ghost-layer-order-and-flux.fixed.md
+++ b/changelog.d/1967-ghost-layer-order-and-flux.fixed.md
@@ -1,0 +1,26 @@
+**The high wall paired its ghost layers backwards, and an inhomogeneous flux offset did not grow
+with the layer.** Two defects in the same loop, in both copies of it.
+
+The high ghosts occupy padded indices `-g .. -1` and `-g` is the one adjacent to the wall, so both
+the ghost walk and the interior walk must start there. The high-wall loop started its ghost walk
+at `-1`, the far end, while its interior walk started near — pairing the layers in reverse for
+`ghost_depth >= 2`. Measured on `cos(2πx)`, which is even about both walls so `NEUMANN(0)` is
+exact at both: the low wall was machine-zero at every depth while the high wall was **5.4e-01 at
+`g=2` and 1.3e+00 at `g=3`**, with `reversed(got) == want` — every value correct, every slot wrong.
+
+Separately, ghost layer `k` sits `(2k-1)·dx` from its mirror, so an inhomogeneous Neumann offset
+must scale with the layer. `dx · v` was applied to every layer — the `k = 1` value used
+throughout, drifting by `(2k-2)·dx·v`.
+
+**Both are invisible at `ghost_depth = 1`**, where a single layer has no order to reverse and
+`(2·1−1)·dx = dx`. Every caller in the library passes 1 or omits it, which is why a suite of 6147
+was green with both present. The one production consumer at depth 3 is `hjb_weno.py:330`, whose
+`_SUPPORTED_BC_TYPES` is `{NEUMANN, NO_FLUX, PERIODIC}` — exactly the family the first defect hits.
+
+Eighteen of twenty-four probed cells are unchanged, every `ghost_depth = 1` cell among them, and
+that is pinned: since every caller uses depth 1, a change there would be a regression in
+everything that currently works.
+
+Found by four independent verification oracles — convergence order, invariance,
+cross-implementation and grid convention — run in parallel on #1966, which had certified this
+family as correct after measuring one wall.
--- a/mfgarchon/geometry/boundary/applicator_fdm.py
+++ b/mfgarchon/geometry/boundary/applicator_fdm.py
@@ -1164,17 +1164,38 @@ class PreallocatedGhostBuffer:
                     lo_interior[axis] = g + k  # Adjacent interior cells from g up
                     buf[tuple(lo_ghost)] = buf[tuple(lo_interior)]
                     if apply_flux:
-                        buf[tuple(lo_ghost)] += dx * v  # Issue #1262: was -= (du/dx sign), now += (du/dn sign)
+                        # #1967: the offset is the mirror SEPARATION times the flux, and that
+                        # separation grows with the layer. Ghost layer k and its mirror interior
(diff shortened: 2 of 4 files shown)


derrring:
## Re-review: MERGE-OK

Independent re-review at `163e0bb1` in an isolated worktree. Both blockers closed by measurement.

**Blocker 1.** The reviewer's exact mutation (high-wall offset reverted to `dx*v` in both copies, per-face guarded on `side == "max"`), full suite: **3 failed, 6191 passed** — previously all green. Four separate mutations, kills within `tests/unit/test_geometry/boundary/` (264 tests):

| mutation | kills | what kills it |
|---|---:|---|
| uniform-low | 6 | 4 exact-continuation + 2 cross-path |
| uniform-high | 6 | 4 exact-continuation + 2 cross-path |
| per-face-low | 3 | cross-path agreement only |
| per-face-high | 3 | cross-path agreement only |

**Blocker 2.** Instrumented call counts through a real `HJBWenoSolver`, depth 3, order 5, confirm "reachable but not currently reached" exactly.

**The 2-D test is not vacuous.** Against `origin/main` it fails at g=2 and g=3 and passes at g=1 — correct, depth 1 is the blind spot. A corner-only mutation (each axis sweep restricted to `slice(g, -g)` on the other axes, every 1-D edge left exact) kills it at g=1,2,3 and **nothing else**: 261 other boundary tests stay green. It is the only corner guard in the directory.

---

## Two findings addressed before merge

**Finding 1 — the superseded blocker-2 claim survived verbatim in the test module docstring** (`test_ghost_layer_order_and_flux_1967.py:14-17`), which is the copy that ships next to the code. The correction had landed only in the changelog fragment. Fixed in `d9a56...`: struck rather than rewritten, with the measured routing table replacing it. This PR body carries the same stale sentence and is superseded by that docstring.

**Finding 2 — the discrimination baseline is stale, and it is not this PR's to fix.** Measured independently:

| tree | `neumann_low_wall_flux_sign` |
|:---|---:|
| recorded baseline | 23 |
| `origin/main` | **30** |
| this PR head | **36** |

`main` is 7 above its own record before this PR touches anything, so the weekly sweep is already red. Re-recording here would absorb `main`'s drift into a bug-fix commit and attribute 13 to a change responsible for 6. Filed as **#1974**, which also notes that the other 23 mutations are unmeasured and that regeneration needs ~50 min of exclusive use of the checkout.

## Remaining non-blocking, carried forward

- Per-face inhomogeneous flux still has no independent oracle — both its mutations are killed only by agreement against the uniform path, which is itself now pinned by an exact continuation. Sound chain, but the `wall` parametrisation went onto the uniform path only.
- Depths ≥ 4 are unpinned: a shared error exact for k ≤ 3 and wrong beyond kills **0** of 264 boundary tests. No production exposure (WENO uses 3), but `pad_array_with_ghosts` is public and takes an arbitrary `ghost_depth`.
- "unreachable today" for the third copy overstates it — `create_ghost_buffer_from_bc` is a public export with no *library* caller. Unreached, not unreachable. That third copy is now fixed in #1973.

## Gate

`./scripts/local_ci.sh` → **6194 passed, 12 skipped, 21 xfailed, GATE GREEN**, 128.9 s.

One earlier run of the same tree went red on `test_every_mutation_anchor_still_matches_exactly_once[4]` and did not reproduce; every one of the 24 anchors counts exactly 1 before and after, and `_SCRIPT` is `Path(__file__).resolve()`-based so it is not CWD-dependent. No mechanism identified and the failure text was not retained — recorded here so a recurrence is cheap to catch, not explained.

Issue #1968, opened 2026-08-16
[Design] PreallocatedGhostBuffer holds two geometries selected by order, derives its spacing with the other one's formula, and its GridType parameter is unreachable


derrring:
One class, three inconsistencies about what its grid is. None is a wrong number on its own; together they are why every measurement of the ghost cells has to state which geometry it assumed, and why two independent oracles reached opposite verdicts on the same code this week.

Found by four parallel verification oracles on #1966/#1967. All **pre-existing**; all measured.

## 1. Two geometries, selected by `order`

| branch | `order` | geometry | exactness |
|---|---|---|---|
| `_apply_linear_reflection` | ≤ 2 | **cell-centred** | exact vs a cell oracle; `8.6e-02` vs a node oracle |
| `_apply_poly_extrapolation` | > 2 | **node-centred** | `1.8e-15` vs a node oracle; `3.8e-01` vs a cell oracle |

Each branch's zero is the positive control for the other's error. Confirmed by reading: `_extrapolate_boundary_1d` uses `interior_indices = range(g+1, g+1+n_stencil)`, **skipping** padded index `g` because that index *is* the boundary node, and places ghosts at `x = -g·dx … -dx`. That is node-centring, unconditional, with no `grid_type` branch. The `order ≤ 2` path meanwhile satisfies `u_b = (u_ghost + u_interior)/2`, the cell-centred relation.

So the same object answers the same question two different ways depending on a numerical-order parameter, and neither branch says which geometry it is using.

## 2. The spacing is computed with the node formula inside the cell-centred class

`applicator_fdm.py:972` derives `_grid_spacing` from `domain_bounds` as `extent / (n_points - 1)` — the node-centred formula.

```
domain_bounds=[[0,1]], interior_shape=(9,)  ->  0.125     (node 1/8; cell would be 1/9 = 0.1111)
domain_bounds=[[0,1]], interior_shape=(8,)  ->  0.142857  (node 1/7; cell would be 1/8 = 0.125)
```

The `order ≤ 2` ghosts are then computed cell-centred against a node-centred `h`.

## 3. `GridType.VERTEX_CENTERED` is unreachable, and inert even when forced

`pad_array_with_ghosts` has no `config` parameter, so `PreallocatedGhostBuffer` always receives `GhostCellConfig()` = `CELL_CENTERED`. Passing one directly changes nothing:

```
[...]

Issue #624, opened 2026-01-22
feat: Extend Adjoint-Consistent Boundary Conditions to n-Dimensions


derrring:
## Summary

As a follow-up to the successful 1D implementation in #574, this issue tracks the work to extend the adjoint-consistent boundary condition framework to handle n-dimensional (2D, 3D) problems.

This is critical for ensuring numerical accuracy and stability for problems where the potentials minimum ("stall point") lies on the domain boundary, a common scenario in many MFG applications.

### 1. Boundary Coupling Module (`mfg_pde/geometry/boundary/bc_coupling.py`)

- Create a new function `create_adjoint_consistent_bc_nd(m_current, geometry, sigma, regularization)`.
- This function will take the current density `m_current`, the `geometry` object, and the diffusion coefficient `sigma`.
- **Gradient Computation**: It should use `geometry.get_gradient_operator()` to obtain a gradient operator for `ln(m)`. This avoids dimension-specific finite difference code and leverages the geometry abstraction.
- **Normal Vector Computation**: It needs to obtain the outward normal vectors at all boundary points. This may require ensuring the `geometry` object provides a method like `get_outward_normals()`.
- **Normal Gradient**: Compute the dot product of the log-density gradient and the outward normals to get `∂ln(m)/∂n` at each boundary point.
- **Robin BC Creation**: Construct a `BoundaryConditions` object containing `BCSegment`s of type `ROBIN`.
    - `alpha` should be 0.0.
    - `beta` should be 1.0.
    - The `value` for the Robin condition (`g` in `αu + β∂u/∂n = g`) will be an array of values, `-σ²/2 * ∂ln(m)/∂n`, computed for each boundary point. This may require verifying/enhancing `BCSegment` and the applicator to handle array-like values.

### 2. HJB FDM Solver Integration (`mfg_pde/alg/numerical/hjb_solvers/hjb_fdm.py`)

- In the `solve_hjb_system` method, modify the `else` block (for `self.dimension > 1`).
- Add a condition to check if `self.bc_mode == adjoint_consistent`.
- If it is, call the new `create_adjoint_consistent_bc_nd` function to generate the dynamic `BoundaryConditions` object for the current time step.
- This new `bc` object should then be passed to the `_solve_single_timestep` method, which in turn will use it when calling `self.bc_applicator.enforce_values`. The `enforce_values` method may need to be updated to handle array-based `value` attributes in `BCSegment`s for Robin conditions.

## Acceptance Criteria

[...]

Issue #1632, opened 2026-07-18
Implement interface_velocity in gradient_centered / gradient_upwind / divergence_centered assembly handlers


derrring:
Follow-up to #1631, which established that only `divergence_upwind` consumes `interface_velocity`. The other three nD FDM advection schemes accept the parameter and ignore it:

| module | `interface_velocity` refs | consumes it |
|---|---|---|
| `fp_fdm_alg_divergence_upwind.py` | 8 | yes (branches at :156, :176) |
| `fp_fdm_alg_gradient_centered.py` | 2 | no (signature + docstring) |
| `fp_fdm_alg_gradient_upwind.py` | 2 | no |
| `fp_fdm_alg_divergence_centered.py` | 2 | no |

#1631 closes the *honesty* gap by rejecting the combination. This issue is the *capability* gap: make the velocity channel work for the remaining schemes, then extend `_INTERFACE_VELOCITY_SCHEMES` in `fp_fdm_time_stepping.py` accordingly.

Why it matters: the velocity channel is how a non-quadratic / MAXIMIZE / regularized Hamiltonian supplies its α\* (#1542, RFC #1574 Phase 1, #1528). Restricting it to one scheme means those Hamiltonians cannot use centered or gradient-form advection at all.

## Notes for the implementer

- `divergence_upwind` is the reference: it reads `interface_velocity[d][multi_idx]` for the right face and `[d][multi_idx_minus]` for the left, replacing `-c*(u_plus - u_center)/dx`.
- The gradient-form schemes need the node-centered equivalent, not a face value — decide the convention explicitly and pin it, rather than reusing the face indexing.
- Each scheme owes a test in the shape of `test_consuming_scheme_actually_honors_the_velocity` (`tests/unit/test_alg/test_fp_fdm_velocity_channel_1528.py`): a velocity differing by O(1) must change the density. Bit-identical output is the bug signature.
- `_INTERFACE_VELOCITY_SCHEMES` is the single source for the accept-list; adding a scheme there without a real branch is caught by the `_COEFFICIENT_NOT_APPLICABLE` NaN tripwire plus the per-timestep NaN guard (#881).

derrring:
## The documented fallback is wrong: the drift is zeroed, not re-derived from U

Two places state that the three non-consuming schemes fall back to `-c*grad(U)`:

- `fp_fdm_time_stepping.py:951` — *"the others fall back to `-c*grad(U)` and still need it (#1632)"*
[...]

PR #1970, opened 2026-08-16
feat(geometry): a provider for the outward normal drift at a no-flux wall

derrring:
Step 2b of the BC consolidation, and the step where the numbers change. Builds on #1965.

## What it is

An impermeable wall is `J · n = 0` with `J = m·v − D·∇m`, i.e. `v_n·m − D·∂_n m = 0` — **Robin in `m` with `alpha = v_n`**. Imposing `∂_n m = 0` instead is the same condition only when the drift is tangential at the wall.

Measured on this package earlier, 1-D, `σ = 0.3`, wall-normal drift 3.2:

```
non-conservative assembly   loses 5.4% of the mass
conservative assembly       machine precision
```

That gap is the drift never reaching the wall. `NormalDriftProvider` is what carries it.

## Why a provider, and why on `alpha`

`v_n = −c·∂_n U` is a functional of the **coupled** solution, so it is knowable only per Picard iterate. It sits on `alpha` and not on `value` because `value` is the homogeneous right-hand side, which for an impermeable wall is zero — and `alpha` only became reachable by a provider in #1965.

## Verified against the flux, not against another code path

Four value functions × two walls through `pad_array_with_ghosts`: **|J·n| ≤ 2.2e−16**.

The **parabolic** row is the discriminating one: there the two walls see drifts of different *magnitude* (0.05 and −0.95) rather than one number negated, so an implementation that computed a single derivative and flipped its sign fails. On a linear field it would pass.

A control asserts the pointwise `∂_n m = 0` wall **leaks** on the same configuration — without it the whole file passes over a provider that silently returns zero, and that mutation is measured below.

## Conventions, and no defaults

- **Outward normal**, matching `BCSegment.beta` and every ghost formula since #1907. Verified on rising, falling and flat fields, since a flipped convention passes a one-signed test.
- **`c` has no default.** It is the Hamiltonian's control law (#1420); a defaulted one silently rescales the wall's drift, and the wrong answer still converges.
- Missing state keys are named individually.
- An **nD wall raises** rather than guessing an axis — guessing would reintroduce exactly the axis-versus-outward-normal confusion #1907 removed. Same limit `AdjointConsistentProvider` states (#624).

## Mutations

```
low-wall sign flipped          3 failed / 16 passed
high-wall sign flipped         2 failed / 17 passed
low wall reads the far end     1 failed / 18 passed
provider returns zero          4 failed / 15 passed   <- the silent-regression case
```

Source restored byte-identical after each.

## Scope

Does **not** wire this into any solver. Step 2c routes the applicators through the resolver and deletes the dispatch chains; that step is currently blocked on #1968 (the class holds two geometries) and on defining what a field-valued coefficient is indexed by.

## Gate

`./scripts/local_ci.sh` → **6178 passed, 12 skipped, 21 xfailed, GATE GREEN**.


derrring:
## Review: MERGE-BLOCKED — and blocked on something larger

Independent adversarial review at `f001afc2`, plus a follow-up trace. Taking the good news first, because it should not be relitigated: **the sign convention is correct**, at both walls, confirmed independently against `AdjointConsistentProvider` by feeding both $U = \ln f$ with $c = -1$ so each computes $\partial_n \ln f$ — they agree to 12 digits at both walls. Every number in the PR body that could be checked reproduces exactly, and the gate result is real.

What is wrong is everything offered as *evidence* that the convention is correct, plus three defects in behaviour — and underneath all of them, the class has nowhere to plug in.

### The reason this cannot be fixed in place: #1975

`NormalDriftProvider` supplies the `alpha` of a **ROBIN** Fokker-Planck wall. Measured across every solver's `_SUPPORTED_BC_TYPES`: **zero of seven FP solvers accept ROBIN.** The only solver in the library that does is `HJBGFDMSolver` — on the HJB side, which is exactly the side where the condition is *not* Robin.

Attempting an end-to-end solve with this PR's own documented segment type:

```
NotImplementedError: HJBFDMSolver does not support boundary condition
type(s) ['ROBIN'] (supported: ['DIRICHLET', 'NEUMANN', 'NO_FLUX', 'PERIODIC'])
```

Refused at construction, before any solve. Filed as **#1975**.

That is also the reason for blocker 1 below: with no solver to run it through, the test had to drive `ghost_cell_robin` directly, and a test that drives the solver of an equation and then checks that equation is an identity.

### Blockers

**1. The headline flux test is an algebraic identity.** It computes `v_n*(ghost+interior)/2 - D*(ghost-interior)/dx` with the *same resolved alpha* read back off the segment — the LHS−RHS of the equation `ghost_cell_robin` just solved. Residual against an arbitrary alpha, provider bypassed:

```
alpha=-1e+06  ->  'J.n' = -2.7350e-11        alpha=0          ->  0.0000e+00
alpha=  -3.7  ->  'J.n' =  1.1102e-16        alpha=1.235e+04  ->  7.7183e-13
```

Its only sensitivity to `alpha` is floating-point conditioning. Confirmed from the other side by ten mutations of `compute`'s arithmetic — **not one of the 8 flux rows failed in any of them**, including *both walls read one cell further in*, *2nd-order one-sided stencil*, and *`dx` hardcoded, geometry ignored*. The PR body's claim that "an implementation that computed a single derivative and flipped its sign fails" the parabolic row is false; every parabolic flux row passes under it.

**2. On a time-dependent `U` it reads the terminal cost, which is fixed data.** `u[-1, :]` selects row `-1`, and `fixed_point_utils.py:393` is `U[-1] = U_terminal`, re-imposed unconditionally every Picard step. Measured here directly, two states differing in **every** non-terminal row:

```
iterate A (scale=1)    v_n = 2.000000000000
iterate B (scale=999)  v_n = 2.000000000000
identical across iterates            : True
equals +c*dU/dx of the TERMINAL cost : True
```

The rows that actually move would give `0.01` and `9.99`. So $v_n$ is a functional of $g(x)$ alone and is identical on every iterate — the exact negation of the PR's stated reason for the class to exist. The comment `# final time slice, matching AdjointConsistentProvider` is where it went wrong: for $m$ the final slice is solution-dependent (FP runs forward), for $U$ it is the terminal condition (HJB runs backward). Copying the slicing across the duality copied a category error.

**3. An nD wall does not raise for `x_min`/`x_max` — it returns the wrong axis.** The guard is `if self.side not in ("left", "right")`, which never consults `geometry.dimension`, and `x_min`/`x_max` alias to `left`/`right`. On a 2-D grid with `U = 3x + 100y`:

```
side=x_min  -> RETURNED 100.0    (exact v_n = +3.0)
side=x_max  -> RETURNED -100.0   (exact v_n = -3.0)
side=y_min  -> raised NotImplementedError
```

`u.ndim > 1` collapses to the last x-row, then differences along **y**. Both the reduction and the axis are wrong, silently. The PR body asserts the opposite: *"An nD wall raises rather than guessing an axis."* (The hole is shared with `AdjointConsistentProvider` and pre-existing there — but this class drops even the `getattr(geometry, "dimension", 1)` read its sibling has.)

**4. The class docstring's only `Example` is the leaking configuration.** Same field, same provider, only `bc_type` varied:

```
bc_type=NO_FLUX : resolved alpha=2.0000  ghost= 0.80000000  J.n = 1.600e+00
bc_type=ROBIN   : resolved alpha=2.0000  ghost=-2.10909091  J.n = 2.220e-16
```

Under `NO_FLUX` the provider resolves correctly and the padding path then ignores `alpha` — `J·n = 1.6`, precisely the leak this PR exists to remove. Only `ROBIN` works, which the docstring never says. The documented call is the bug.

### Non-blocking, most severe first

1. **The wall condition is imposed to first order and nobody measured it.** True $J\cdot n$ over 7 refinements: `4.88e-02 → 1.04e-02 → 2.53e-03 → 6.27e-04`, order **1.00**. Defensible — the *ghost* perturbation is $O(dx^2)$ because $\delta\text{ghost} \approx m_0\,\delta\alpha\,dx/D$ damps the coefficient error by one power — but the PR reports `2.2e-16` where the physical residual at $dx = 0.05$ is $4.9\times10^{-2}$.
2. **`'drift_coefficient'` is written to the state by nothing in the library.** `fixed_point_iterator.py:605` builds `{m_current, U_current, geometry, sigma, iteration}`. Positive control: the same grep form for `"sigma":` returns six library writers. So `NormalDriftProvider(side)` without an explicit `c` raises `KeyError` in the real Picard loop. The docstring asserts a wiring that does not exist.
3. **Second copy of the drift convention.** The library's single source for FP advection velocity is `compute_fp_velocity_field` (`fixed_point_utils.py:397`), which evaluates `H_class.optimal_control(...)`. This hardcodes $v = -c\nabla U$, valid only for quadratic-separable-MINIMIZE.
4. **`float()` and `-> float` re-narrow what #1965 widened.** #1965 removed a `float()` one layer up precisely so a provider could return a field, against a protocol declaring `float | NDArray[np.floating]`. Harmless in 1-D where the wall is a point; structurally wrong for the 2-D wall that is step 2c's target, where `alpha` varies along the boundary.
5. **The PR body's fourth mutation is mislabelled.** "provider returns zero → 4 failed / 15 passed" reproduces as **5 failed / 14** for the whole provider; 4/15 is the *low wall only*.
6. **Error ordering masks the diagnostic.** The `side` check sits after the `c` lookup, so `NormalDriftProvider("y_min")` reports the `drift_coefficient` `KeyError` rather than `NotImplementedError`.
7. **The "5.4% of the mass" figure has no locatable artifact**, and is now asserted in three shipped locations including a library docstring.

### Recommendation

Hold, not iterate. Blockers 2, 3 and 4 are each a small fix, but blocker 1 cannot be fixed while #1975 stands — there is no solver to build a real oracle against. The oracle this area needs is **mass conservation against `FPParticleSolver`**, which already imposes the correct wall via Skorokhod reflection; that becomes available once a grid FP solver accepts ROBIN.

Marking draft pending #1975.

derrring:
## Independent literature corroboration for this PR's premise

A systematic read of the MFG literature today (filed as #1992) reached this PR's central claim from the source side, without knowing the PR existed.

**Bardi & Cirant**, in *PDE Models for Multi-Agent Phenomena* (Springer INdAM 28, 2018), pp. 1–2, system (1.2), read from a **rendered page** rather than the text layer — that volume's text layer drops every `Δ` and `Ω`:

```
∂_n v_k = 0                                  (HJB side)
∂_n m_k + m_k D_p H_k(x, D v_k) · n = 0      (FP side)
```

The FP wall condition carries `D_p H`, exactly as this PR argues. `∂_n m = 0` is the special case where the drift is tangential at the wall — which is what the PR says and what the 5.4% mass loss measures.

Two things this adds beyond confirmation:

1. **The literature calls both lines "Neumann"**, which is harmless in prose and fatal once a *name* drives a dispatch. Cirant's own paper title says Neumann while its `k_n)` line is Robin. So the naming hazard this PR navigates is real and pre-existing in the sources, not an artefact of our BC layer.

2. **This is not only a boundary-condition question — it is the same defect as #1632, at the wall.** #1632 is that a non-separable Hamiltonian's precomputed `alpha* = D_p H` gets silently discarded in the interior on three of four advection schemes. This PR is that `D_p H` never reaches the wall row. Same quantity, two places it fails to arrive. Worth fixing with that connection in view: an interior fix that leaves the wall on `∂_n m = 0` still solves a different problem near the boundary, and the conservative-vs-non-conservative gap this PR measures would partly mask it.

Status: branch is 14 commits behind `main` but still `MERGEABLE / CLEAN`. The premise is sound and now has an external source behind it; what it needs is a rebase and a re-derivation of whatever the #1975 rewrite invalidated.

derrring:
## Traced the whole chain; it breaks at the last link, not where I said earlier

I said the provider had "no production call site". That was the symptom, not the cause. Measured, link by link:

| link | state |
|---|---|
| provider sits on `BCSegment.alpha` | yes — and correct, since `value` is the homogeneous RHS, zero for an impermeable wall |
| `has_providers()` recognises `alpha` | yes — docstring says "in `value`, `alpha` or `beta`" |
| `with_resolved_providers()` resolves `alpha` | yes — it loops over `("value", "alpha", "beta")` |
| `using_resolved_bc` runs per Picard step | yes — `fixed_point_iterator.py:623` |
| **FP assembly reads `segment.alpha`** | **no — zero occurrences in `fp_solvers/`** |

So the provider is placed correctly, the live resolution path does handle `alpha`, and the resolved wall-normal drift arrives at a door nobody opens. The assembly takes `coupling_coefficient` (a scalar) plus `u_flat` and re-derives the velocity as `-coupling_coefficient * grad(U)` itself — the separable-quadratic special case, and the same limitation #1632 just fixed in the interior.

**Correcting myself twice:** the premise did not die with #1975 — this branch rebases onto main with 0 conflicts, its 19 tests pass, and the wider geometry suite passes 1254. And "no production call site" pointed at the wrong link.

## On merging

I do not think this should merge as-is. It is a correct part built against the resolver architecture in `geometry/boundary/resolution.py`, and that architecture has **zero production consumers**: `ResolvedBC` is produced, exported, and never read, and `resolve_bc` has no caller outside its own inline smoke test. Filed as #2005.

Merging would add a second correct-but-unwired part beside the first, which is a shape this repository keeps paying for — `high_order_ghost_neumann` was also built, never wired, and never verified until someone measured it.

Better sequencing: #2005 first, deciding which side owns the wall assembly, then this provider lands into something that consumes it. The measurement here — 1-D, sigma = 0.3, wall-normal drift 3.2, non-conservative assembly losing 5.4% of the mass against machine precision for the conservative one — is exactly the evidence that makes #2005 worth doing. It is not wasted work; it is the motivating measurement filed before the thing it motivates.

Bardi and Cirant (INdAM 28, equation 1.2, read from a rendered page) confirm the physics independently. The library's own `FPResolver` docstring states it correctly too. Neither is reachable from a solver today.

derrring:
Converting to draft and marking blocked on #1979, rather than closing.

The code here is correct and tested. What it lacks is a working consumer, and #1979's body names this
PR directly:

> A user who wires a coupled wall coefficient — the whole point of the `BCValueProvider` layer, and
> what `AdjointConsistentProvider` and **#1970's `NormalDriftProvider`** exist to supply — gets a
> **no-flux wall and no diagnostic** on the FDM path.

Measured there:

```
FEM   assemble_robin_terms    -> TypeError: float() argument ... not 'AdjointConsistentProvider'
FDM   solve_timestep_full_nd  -> NO ERROR; silently assembled, byte-identical to no-flux
```

So merging this today ships a class whose only use is either silently dropped or a `TypeError`. That
is worse than not shipping it, because a public provider reads as a capability — the same shape as
the 19 solvers that declare `honors_inhomogeneous_neumann` without anyone having made that claim
(#1980's census).

Unblocks when `_BOUNDARY_HANDLERS` receives `boundary_conditions` and reads `alpha` — #1979, which is
`size: small` and already prioritised.

derrring:
Out of draft, restacked on #2013 (which it was blocked pending), gate GREEN.

The two interact directly now: **the FDM grid paths refuse a provider-valued wall coefficient**, so
this provider's only working consumer is the weak-form path.

That refusal is the point rather than a limitation to route around. This provider exists to stop a
wall silently reverting to `∂_n m = 0`; being *dropped* by the FDM handlers — which take no
`boundary_conditions` argument at all — would have delivered exactly that under a different name. So
the blocking order was right: shipping the provider before the guard would have created a
silent-wrong path, not merely an unguarded one.

Three things added:

- The class docstring now says **where it works** (`FPFEMSolver` reads `alpha`/`beta`/`g` in its weak
  form) and that `bc.with_resolved_providers(state)` is a **downgrade**, not a workaround — this
  provider is recomputed each Picard iterate from the current state, and a frozen normal drift is a
  constant Robin coefficient, i.e. a different boundary condition.
- A pin that the FDM path refuses it, with a float control so the refusal cannot be a blanket one.
- A pin that the downgrade claim is **measurable rather than rhetorical**: `compute` reads
  `U_current`, so two value functions give two wall coefficients while a resolved provider gives one
  for both.

One correction to my own work, worth stating because the failure mode is the one this repo keeps
hitting: the first version of those tests was written against a **guessed** API —
`NormalDriftProvider(sigma=...)` and `.value(state)` — where the real one is
`(side, drift_coefficient=None)` and `.compute(state)`. Citing a signature is claiming it exists.
They were rewritten after reading it.

Issue #1972, opened 2026-08-16
ghost_cell_neumann's cell-centred branch is off by a factor of 2, and the calculator path reads the BC value with the opposite sign convention to the live one


derrring:
Found while writing the pin for #1971. Same route, different defect — filed separately rather than bundled.

## The oracle

`u = 3x` is exact for any first-order Neumann ghost rule, so the correct ghost is just `u` at the ghost centre. `h = 0.125`. Each grid type is compared against **its own** ghost geometry (cell-centred: interior at `+h/2`, ghost at `-h/2`; vertex-centred: interior on the wall, ghost at `-h`) — the separation is `h` in both, so a Neumann rule has no geometric reason to branch on grid type at all.

`flux_value` admits two readings and they differ only at the min wall, so that wall discriminates: **outward** = `du/dn` (`-3` at min, `+3` at max), **axis** = `du/dx` (`+3` at both).

## Measured — `ghost_cell_neumann` (`ghost_cells.py:80`)

```
grid type         wall  reading           got       want         err
CELL_CENTERED     min   outward       0.93750   -0.18750  +1.125e+00
CELL_CENTERED     min   axis         -0.56250   -0.18750  -3.750e-01
CELL_CENTERED     max   outward       3.56250    3.18750  +3.750e-01
CELL_CENTERED     max   axis          3.56250    3.18750  +3.750e-01
VERTEX_CENTERED   min   outward       0.37500   -0.37500  +7.500e-01
VERTEX_CENTERED   min   axis         -0.37500   -0.37500  +0.000e+00
VERTEX_CENTERED   max   outward       3.37500    3.37500  +0.000e+00
VERTEX_CENTERED   max   axis          3.37500    3.37500  +0.000e+00
```

Two separable findings.

### 1. The cell-centred branch has a spurious factor of 2 — wrong under either reading

```python
if grid_type == GridType.VERTEX_CENTERED:
    return interior_value + dx * flux_value * outward_normal_sign
else:
    return interior_value + 2.0 * dx * flux_value * outward_normal_sign
```

The error is `±0.375 = h*C` at every cell-centred row, i.e. exactly one extra `dx*g`. The docstring asserts the geometry that makes it wrong:

```
du/dn = (u_ghost - u_interior) / (2*dx) * sign = g
```

[...]

PR #1973, opened 2026-08-16
fix: pair GhostBuffer's ghost and interior slices in the same direction (#1971)

### LINKED ISSUE #1971: GhostBuffer._update_bounded pairs by slice and reverses both walls (one of five copies, not the last)
`GhostBuffer._update_bounded` (`applicator_fdm.py:756-790`) pairs ghost and interior **by slice**, mapping the outermost ghost onto the wall-adjacent interior. It is the third copy of the arithmetic corrected in #1967, and unlike the other two it reverses **both** walls.

Found by independent review of the #1967 fix. **Pre-existing.** The #1967 PR body says "the loop has two copies" — that is false about the file, and this issue is the correction.

## Measured

Same oracle as #1967: `u = cos(2πx)` is even about both `x = 0` and `x = 1`, so `NEUMANN(0)` is exact at both walls and the correct ghost is the field itself at the ghost centres.

```
g=1   low 0.000e+00                       high 1.110e-16
g=2   low 5.412e-01  reversed match True  high 5.412e-01  reversed match True
g=3   low 1.307e+00  reversed match True  high 1.307e+00  reversed match True
```

Identical magnitudes to the defect fixed in #1967, identical signature — `reversed(got) == want`, every value correct and every slot wrong — and present at both walls rather than one.

## Reachability

`create_ghost_buffer_from_bc` has **no library caller** outside the lazy-import registry in `geometry/boundary/__init__.py`. So this is latent: the "public API that only fires when a user hits it" class, not a live defect. That is why it is filed rather than folded into #1967, whose scope was the two reachable copies.

## Why three copies

The same per-layer pairing is written out in:

| site | walls affected | status |
|---|---|---|
| `PreallocatedGhostBuffer._apply_linear_reflection` | high | fixed, #1967 |
| `PreallocatedGhostBuffer._apply_ghost_for_face` | high | fixed, #1967 |
| `GhostBuffer._update_bounded` | **both** | this issue |

A fourth site, `_apply_ghost_for_face`'s terminal `else` (`:1804-1813`), still carries the pre-fix `-(k+1)` against `-(2g+k+1)`. It is unreachable because all eight `BCType` members are handled above it, but any new member would inherit the defect.

Corroboration that the corrected form matches the file's own convention: `-(g - k)` was **already** the high-ghost index in the `ROBIN` (`:1231`, `:1763`) and `EXTRAPOLATION` (`:1270`, `:1800`) branches. The `NEUMANN` branch was the outlier, in every copy.

## What a fix owes

The oracle of #1967 — an exact continuation at the ghost centres, asserted at **both** walls and at depths 1 through 3 — plus a control that depth 1 does not move. `_update_bounded` takes a `Calculator`, so the fix is a pairing change and not an arithmetic one; the values it computes are already right.

Related: #1967 (the two reachable copies), #1966, #1968.

derrring:
Closes #1971.

The copy in `GhostBuffer._update_bounded`, and the only one of the three *mirror* copies that reverses **both** walls.

> **[CORRECTED]** This PR originally said "third and last copy". That is false: the same slice-pair construct is written out twice more in `PreallocatedGhostBuffer` (`applicator_fdm.py:1120-1135` `_apply_linear_reflection` DIRICHLET, and `:1673-1687` `_apply_ghost_for_face`), both reversed at **both** walls, `reversed(got) == want`, magnitudes 1.110 / 2.674 / 3.414 at g = 2..4. Those are #1966 Defect 2 and are not fixed here. `_update_ghosts_legacy` carries it too. See the re-review comment below.

## The defect

It differs in kind from the other two. Those were per-layer loops with a wrong index expression; this one is a vectorised slice assignment where the two slices run opposite ways.

```
lo_ghost    = [0:g]     index g-1 adjacent to the wall, index 0 outermost
lo_interior = [g:2g]    index g   adjacent to the wall
buf[lo_ghost] = calc(buf[lo_interior])    -> outermost ghost gets the nearest interior
```

The high side has the mirror problem: its ghost slice runs nearest-first, its interior slice `[-2g:-g]` runs farthest-first.

**At `ghost_depth = 1` a one-element slice is its own reverse**, which is why the only depth anything uses looked correct.

## Measurement

`u = cos(2*pi*x)`, `N = 8`, `h = 0.125`, `NEUMANN(0)` — even about both walls, so the condition is exact there and the correct ghost is the field itself at the ghost centre.

```
g=1  low_err 0.000e+00   high_err 1.110e-16
g=2  low_err 5.412e-01   high_err 5.412e-01   reversed(got) == want, both walls
g=3  low_err 1.307e+00   high_err 1.307e+00   reversed(got) == want, both walls
```

Same magnitudes as the two copies fixed in #1967 — every value right, every slot wrong.

## The shape of the fix, and why

```python
lo_interior[axis] = slice(2 * g - 1, g - 1, -1)
hi_interior[axis] = slice(-g - 1, -2 * g - 1, -1)
```

It steps the **interior** slice backwards rather than reversing the calculator's output. The output would have to be reversed along `axis`, and the obvious `[::-1]` reverses axis 0 — correct in 1-D and wrong for any other axis in n-D.

That version was written first. It is mutation M3 below, and **it passes every 1-D assertion in the new file.**

## Oracle

Category 1, external: an exact continuation. Not an agreement test, so it survives any later consolidation of these three copies into one owner — which is the direction this campaign is going.

## Mutations

`sha256` of `applicator_fdm.py` verified identical before and after each.

| mutation | result |
|---|---|
| both walls reverted to main's pairing | 13 failed / 8 passed |
| low wall only | 8 failed / 13 passed |
| high wall only | 8 failed / 13 passed |
| `[::-1]` on the output (axis-0 only) | **3 failed / 18 passed** — the two n-D tests only |

The last row is why the 2-D corner test and the non-square-grid test are in the file. Without them this PR would have shipped an axis-0-only fix with a green suite.

## Blast radius: none in the library

`GhostBuffer` — as distinct from `PreallocatedGhostBuffer` — has no caller outside `geometry/boundary/`. The two live consumers, `pad_array_with_ghosts` and `hjb_weno.py:350`, both construct `PreallocatedGhostBuffer`, which does not use the calculators. Reachable only by a user calling the public API.

Same latent class as the #1966 siblings.

## Found on the way, filed separately

**#1972** — `ghost_cell_neumann`'s cell-centred branch is off by a factor of 2 (wrong under either reading of `flux_value`), and the calculator path reads `BCSegment.value` as an **axis** derivative where the live path reads it as an **outward-normal** derivative. Measured against `u = 3x` at both walls and both grid types. Not bundled here.

That defect is why `test_depth_one_is_unchanged` pins the flux types at `value = 0`. **[CORRECTED]** The disagreement is not exhausted by #1972: a third term, the layer-scaled offset, is missing because `Calculator.compute(interior_value, dx, side)` carries no layer index. Measured at `neumann(2.0)`, `h=0.125`: `GhostBuffer` gives a flat `[-0.5, -0.5, -0.5, -0.5]` where the required ratio is `[1, 3, 5, 7]`. Added to #1972.

## Gate

`./scripts/local_ci.sh` → **6246 passed, 12 skipped, 21 xfailed, GATE GREEN**, 117 s.

`uv.lock` was dirty in the working tree again (second occurrence) and was restored to `main`'s before committing — the gate never invokes uv, so it would have gone in unmeasured.

### DIFF
--- a/changelog.d/1971-ghost-buffer-layer-pairing.fixed.md
+++ b/changelog.d/1971-ghost-buffer-layer-pairing.fixed.md
@@ -1,6 +1,9 @@
-`GhostBuffer` paired ghost and interior layers in opposite directions at `ghost_depth > 1`, on
-both walls — the third and last copy of the reversal corrected in #1967. The ghost slice `[0:g]`
-runs outermost-first while `[g:2g]` runs nearest-first, so the element-wise assignment gave ghost
-layer 1 the value computed for the farthest interior cell; the high side had the mirror problem.
-At `ghost_depth = 1` a one-element slice is its own reverse, which is why the only depth anything
-uses looked correct. (#1971)
+`GhostBuffer` paired ghost and interior layers in opposite directions at `ghost_depth > 1`, on both
+walls. The ghost slice `[0:g]` runs outermost-first while `[g:2g]` runs nearest-first, and the
+element-wise assignment gave ghost layer 1 the value computed for the farthest interior cell; the
+high side had the mirror problem. At `ghost_depth = 1` a one-element slice is its own reverse,
+which is why the only depth anything uses looked correct.
+
+This is the copy in `GhostBuffer._update_bounded` and it does **not** close the family: the same
+slice-pair construct is written out twice more in `PreallocatedGhostBuffer`, both reversed at both
+walls, tracked as #1966 Defect 2. (#1971)
--- a/changelog.d/1971-ghost-buffer-layer-pairing.fixed.md
+++ b/changelog.d/1971-ghost-buffer-layer-pairing.fixed.md
@@ -1,0 +1,6 @@
+`GhostBuffer` paired ghost and interior layers in opposite directions at `ghost_depth > 1`, on
+both walls — the third and last copy of the reversal corrected in #1967. The ghost slice `[0:g]`
+runs outermost-first while `[g:2g]` runs nearest-first, so the element-wise assignment gave ghost
+layer 1 the value computed for the farthest interior cell; the high side had the mirror problem.
+At `ghost_depth = 1` a one-element slice is its own reverse, which is why the only depth anything
+uses looked correct. (#1971)
--- a/mfgarchon/geometry/boundary/applicator_fdm.py
+++ b/mfgarchon/geometry/boundary/applicator_fdm.py
@@ -760,12 +760,20 @@ class GhostBuffer:
             lo_ghost = [slice(None)] * d
             lo_ghost[axis] = slice(0, g)
             lo_interior = [slice(None)] * d
-            lo_interior[axis] = slice(g, 2 * g)
+            # #1971: the ghost slice [0:g] runs OUTERMOST-first (index g-1 is the one adjacent to
+            # the wall) while [g:2g] runs nearest-first, and the assignment pairs them
+            # element-wise -- so ghost layer 1 received the value computed for the FARTHEST
+            # interior cell. Stepping the interior slice backwards makes both run wall-outward,
+            # which is the pairing a mirror means. Reversing the calculator's OUTPUT instead
+            # would have to be done along `axis`, and the obvious `[::-1]` reverses axis 0.
+            lo_interior[axis] = slice(2 * g - 1, g - 1, -1)
 
             hi_ghost = [slice(None)] * d
             hi_ghost[axis] = slice(-g, None)
             hi_interior = [slice(None)] * d
-            hi_interior[axis] = slice(-2 * g, -g)
+            # Mirror problem on this side: the ghost slice runs nearest-first, the interior
+            # [-2g:-g] runs farthest-first.
+            hi_interior[axis] = slice(-g - 1, -2 * g - 1, -1)
 
             # Get interior arrays (views, not copies)
             interior_lo = buf[tuple(lo_interior)]
@@ -786,6 +794,10 @@ class GhostBuffer:
                 **kwargs,
             )
 
+            # At g=1 a one-element slice is its own reverse, which is why every caller saw the
+            # right answer. Measured at g=2 and g=3 on u = cos(2*pi*x), even about both walls so
+            # Neumann(0) is exact there: 5.4e-01 and 1.3e+00 at BOTH walls, with
+            # reversed(got) == want -- every value correct, every slot wrong. (#1971)
             buf[tuple(lo_ghost)] = ghost_lo
             buf[tuple(hi_ghost)] = ghost_hi
 
--- a/tests/unit/test_geometry/boundary/test_ghost_buffer_layer_pairing_1971.py
+++ b/tests/unit/test_geometry/boundary/test_ghost_buffer_layer_pairing_1971.py
@@ -1,8 +1,9 @@
 """`GhostBuffer` paired its ghost and interior slices in opposite directions. #1971
 
-The third and last copy of the pairing corrected in #1967, and the only one that reverses **both**
-walls. It differs in kind from the other two: they were per-layer loops with a wrong index
-expression, this one is a vectorised slice assignment where the two slices run opposite ways.
+The copy of the #1967 pairing defect that lives in `GhostBuffer._update_bounded`, and the only one
+of the three *mirror* copies that reverses **both** walls. It differs in kind from the two fixed in
+#1969: those were per-layer loops with a wrong index expression, this one is a vectorised slice
+assignment where the two slices run opposite ways.
 
     lo_ghost    = [0:g]        index g-1 is adjacent to the wall, index 0 outermost
     lo_interior = [g:2g]       index g   is adjacent to the wall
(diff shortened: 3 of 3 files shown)


derrring:
Rebased onto `main` after #1969 merged (both touch `applicator_fdm.py`, in different regions — no conflict). `./scripts/local_ci.sh` on the rebased head → **6227 passed, 12 skipped, 21 xfailed, GATE GREEN**, 123.4 s.

derrring:
## Both blockers fixed — and the review was right that neither was about the fix

The independent review returned **MERGE-BLOCKED** with two blockers, both false claims I made about the *extent* of the defect family. The one-line fix itself was verified correct against an oracle I did not supply — random interior data plus a first-principles mirror permutation — exact to `0.0e+00` in 1-D at `g=1..6` both walls, 2-D non-square unequal-spacing at `g=1..4` including corners, and 3-D with all extents and spacings distinct at `g=1..4`. All four of my mutation counts reproduced exactly, including the headline: under the `[::-1]` version **all 13 one-dimensional assertions passed** and only the two n-D tests failed.

### Blocker 1 — "third and last copy … the only one that reverses both walls" was false

Two more copies of the same slice-pair construct, both in `PreallocatedGhostBuffer` — the class this PR calls the live one:

```
applicator_fdm.py:1120-1135   _apply_linear_reflection, DIRICHLET branch
applicator_fdm.py:1673-1687   _apply_ghost_for_face, per-face path
```

Verified independently: `reversed(got) == want` at **both** walls, magnitudes `1.110` / `2.674` / `3.414` at `g = 2..4`. `_update_ghosts_legacy` carries it too. These are **#1966 Defect 2**.

The claim was in three shipped places — PR body, changelog fragment, test docstring — and the changelog becomes release-notes copy telling users the family is closed. All three corrected. This is the same error #1969's body made ("the loop has two copies"), which is the reason this PR exists at all; the standard has to apply to it too.

### Blocker 2 — the nonzero-flux disagreement is not exhausted by #1972

There is a **third** term. `Calculator.compute(interior_value, dx, side)` carries no layer index, so the offset stays constant across layers where it must scale as `2k-1` — the second defect #1969 fixed in the siblings, still present here. Measured, `neumann(2.0)`, `h=0.125`, offset of ghost layer `k` from its mirror:

```
GhostBuffer              low  [-0.5, -0.5, -0.5, -0.5]   flat
PreallocatedGhostBuffer  low  [0.25, 0.75, 1.25, 1.75]   = 0.25 * [1, 3, 5, 7]
required ratio                [   1,    3,    5,    7]
```

No constant factor and no sign flip turns a flat sequence into `1:3:5:7`. #1972 covers the factor of 2 and the sign convention and says nothing about layer scaling, so after it lands this route is **still** wrong at `g >= 2`. Recorded in the docstring; being added to #1972.

### The fixture was hiding a wall confusion

`cos(2*pi*x)` on `N=8` is **palindromic on the interior**, so a high-wall/low-end confusion was value-identical on it. It passed 20 of the 21 old assertions, and exactly **one** test in all 1181 of `tests/unit/test_geometry/` caught it — unparametrised. That single test was carrying the entire wall/axis discrimination load.

Replaced with seeded random data and an explicit geometric mirror permutation. The oracle is unchanged in kind: the evenness constraint binds the **ghosts**, not the interior, so the fixture was always free to be anything, and random data has no symmetry to hide behind. Added 3-D, parametrised the axis, extended to `g=6`.

### Mutation table, old file → new file

| mutation | old | new |
|---|---|---|
| revert both walls | 13 F / 8 P | **30 F / 10 P** |
| low wall only | 8 F / 13 P | **27 F / 13 P** |
| high wall only | 8 F / 13 P | **27 F / 13 P** |
| `[::-1]` on output (axis-0 only) | 3 F / 18 P | **8 F / 32 P** |
| **high wall reads the low end** | **1 F / 20 P** | **33 F / 7 P** |
| **exact for `g<=4`, wrong beyond** | **0 F / 21 P** | **6 F / 34 P** |

The last two are the review's coverage findings. Both were effectively uncaught.

### Other corrections accepted

- **`np.flip(out, axis=axis)` on the calculator output is equally correct.** My stated reason for the interior-slice form was sound (`[::-1]` reverses axis 0); the implication that it was the *only* correct route was not. Said so in the docstring.
- **The gate discrepancy is explained.** The body said 6180; the reviewer measured 6227. 6180 was measured **before** the rebase onto #1969 and I noted 6227 in a comment after rebasing but never updated the body. Now **6246** with the expanded file. Body corrected.
- **`_update_periodic` is a translation, not a mirror**, and both slices forward is correct there — confirmed, not touched.

### Gate

`./scripts/local_ci.sh` → **6246 passed, 12 skipped, 21 xfailed, GATE GREEN**, 117 s.

Ready for re-review.

derrring:
## Re-review: MERGE-OK

The corner blocker is closed and verified from the other side. Six corner mutations, the file this replaced vs this one:

| corner mutation | before | after |
|:---|---:|---:|
| filled with a constant | **0 F** | 4 F |
| never written (both sweeps transverse-interior-only) | **0 F** | 4 F |
| swapped with the diagonally opposite corner | **0 F** | 4 F |
| left as the raw interior corner block | **0 F** | 3 F |
| reversed along axis 0 — this PR's own signature defect | **0 F** | 3 F |
| transposed | **0 F** | 3 F |

Zero of six before, six of six now. (3 rather than 4 where the mutation is a no-op at `g=1`.) A NaN-poison probe on the current code finds **0 unwritten cells** of 70 / 108 / 154 / 208 at `g = 1..4`.

`_mirror_index_map` was derived independently two ways — `np.pad(np.arange(n), g, mode="symmetric")` and explicit cell-centre reflection arithmetic — and all three agree for every `n ∈ {1,2,3,4,5,6,8,9,12}` and `g = 1..6`, both walls. No off-by-one. The whole-array assertion reproduces every removed per-axis assertion when sliced the same way, plus 4/16/36/64 newly covered corner cells. All six mutations in the body reproduce to the digit.

## One thing fixed after the re-review (`ee058140`)

The comment I added with the corner oracle overclaimed, and the commit that introduced it **contradicted its own code comment** — the message said "not reachable *that way*" while the comment said the assertion fires "if either axis sweep skips or double-writes a corner block". Both halves are false:

- the axis-0 sweep skipping the corner is **byte-identical** to correct output, because the axis-1 sweep's full-span write repairs it — no observable, nothing to catch;
- the **correct** code double-writes the corner by design, for the same reason, and passes.

Replaced with what it does catch (a corner differing from the separable mirror) and the six-mutation table. Same species as the two blockers this PR already absorbed, which is why it was worth a commit rather than a note.

## A correction to my own review brief

I told the reviewer the commit claimed a never-written corner was "unreachable". The commit said "not reachable **that way**", and the qualifier carries the claim: restricting *one* sweep is unobservable, restricting *both* is reachable, leaves the corner at its `np.zeros` value, and passes the old `isfinite` guard. That case is C2 above and the new oracle catches it. My paraphrase dropped the qualifier; the artifact had it right.

## Gate

`./scripts/local_ci.sh` → **6246 passed, 12 skipped, 21 xfailed, GATE GREEN**, reproduced independently by the reviewer at 137.6 s and by me at 145.6 s.

Ready to merge.

PR #1976, opened 2026-08-16
test(geometry): pin the FP boundary coefficient gap (#1975)

### LINKED ISSUE #1975: [Design] A general Robin FP wall is unavailable on the grid paths, and a provider-valued coefficient fails silently there
> **[REWRITTEN 2026-08-17]** — this issue was filed on a false premise and corrected twice. The original text asserted that no FP solver imposes the Fokker-Planck reflecting wall. **Both halves of that were wrong**: `FPFEMSolver` implements a general Robin and reads the coefficients, and the *default* FDM scheme already imposes $J\cdot n = 0$ structurally. The prescribed fix would have broken two currently-correct paths. Full history at the bottom; the body below is the corrected account.

## What is actually true, measured

`add_boundary_no_flux_entries_conservative` (`fp_fdm_bc.py:222`, aliased at `:396` to `divergence_upwind`) zeroes the **total** face flux $F_{-1/2} = F_{N+1/2} = 0$ — advective plus diffusive. That *is* $J\cdot n = 0$.

$\sigma = 0.3$, wall-normal drift $3.2$, $D = \sigma^2/2 = 0.045$, $n_x = 81$, 200 steps:

| scheme | mass drift | $\partial_n m$ at the wall |
|:---|---:|---:|
| `gradient_centered` | **−78.05 %** | +3.7 |
| `gradient_upwind` | **−75.47 %** | +2.1 |
| `divergence_centered` | **−0.0000 %** | +1962 |
| `divergence_upwind` **(default)** | **−0.0000 %** | +1138 |

Mass conserved to machine precision at a drifted wall, with $\partial_n m$ large and nonzero, is only possible under $J\cdot n = 0$. Positive control — refinement against the exact relation $\partial_n m = (v/D)\,m$:

```
nx= 201   ratio = 0.7377
nx= 401   ratio = 0.8491
nx= 801   ratio = 0.9184
nx=1601   ratio = 0.9574
```

Monotone to 1 at first order. `gradient_upwind` returns exactly `0.0000` — that family really does impose $\partial_n m = 0$, and is already documented non-conservative (#1075).

**So the library's default FP scheme imposes the correct reflecting wall.** It does so *structurally* — by conserving the flux — without any BC-type branch naming the condition. `FPParticleSolver` does the same thing by a different route (Skorokhod reflection of the path). Two of the FP paths are right, and neither says so anywhere a grep can find.

### And a general Robin is implemented too

`FPFEMSolver` overrides `_robin_operator_terms` → `assemble_robin_terms` (`fem/bc_adapter.py:255`), which reads `segment.alpha`, `segment.beta`, `segment.value` and assembles $D(\alpha/\beta)\int_{\partial\Omega}\phi_i\phi_j$ with load $D\beta^{-1}\int_{\partial\Omega} g\,\phi_i$. Measured: $\alpha$ exactly linear (`312.187500` against `999/3.2`), $\beta$ exactly inverse, $g$ into the load. `tests/integration/test_fem_robin_bc.py` passes 13 tests including $O(h^2)$ convergence to a manufactured Robin solution.

## The trap: do not hand the FEM path $\alpha = v_n$

The conservative weak form $M/\mathrm{d}t + DK - C^\top$ already has $\int_{\partial}(J\cdot n)\phi = 0$ as its **natural** boundary condition. Adding $A_{\text{robin}}$ with $\alpha = v_n,\ \beta = -D$ counts $v_n$ a second time, so the implied wall becomes $D\,\partial_n m = 2 v_n m$. Measured on a 1-D rod, $\sigma = 0.3$, drift $3.2$, 40 steps:

```
NO_FLUX (robin hook no-op)        mass drift = -7.4e-15    machine precision
ROBIN(alpha=v_n, beta=-D, g=0)    mass drift = -79.53%     colsum(A_robin) L1 = 6.40 = 2 x 3.2
```

$A_{\text{robin}}$ is not column-sum-free, so it leaks by construction.

**This is why the original prescription — "make the FP boundary assembly read $(\alpha,\beta,g)$" — was wrong.** On a conservative assembly the reflecting wall is already imposed; reading the coefficients and adding their contribution double-counts it.

## The gap that is actually left

Three things, all narrower than what this issue originally claimed.

### 1. A *general* Robin is unavailable on the grid paths

$\alpha \ne v_n$, or $g \ne 0$. The FDM boundary dispatch (`_BOUNDARY_HANDLERS`, `fp_fdm_time_stepping.py:110`) is keyed on the advection scheme and its handlers are not even passed `boundary_conditions` — the parameter is absent from the signature, so no configuration can make them read a coefficient. Measured across 768 combinations (ndim ∈ {1,2,3}, npts ∈ {7,11,21}, σ ∈ {0, 0.05, 0.3, 2.0}, dt ∈ {1e-4, 1e-3, 0.5}, all four schemes, four $(\alpha,\beta,g)$ triples): **zero** non-identical cases against no-flux.

That is a real gap — but it is "a general Robin wall is not expressible on the grid paths", not "the reflecting wall is unimplemented".

### 2. A provider-valued coefficient fails silently on FDM and cryptically on FEM

```
FEM  assemble_robin_terms   -> TypeError: float() argument must be a string or a real number, not 'AdjointConsistentProvider'
FDM  solve_timestep_full_nd -> NO ERROR; silently assembled, byte-identical to no-flux
```

The FDM behaviour is a fail-silent defect: a user who wires a coupled wall coefficient gets a no-flux wall and no diagnostic. The FEM site is `bc_adapter.py:308`, `float(getattr(segment, "alpha", 1.0))` — and the same function raises a carefully worded `NotImplementedError` for `beta == 0` and for a non-constant `g` three lines below. The guard exists and was not applied to `alpha` and `beta`.

### 3. `FPResolver` has zero library callers — and that is now explained

It emits `ROBIN(alpha=v_n, beta=-D, g=0)` for an impermeable wall. Nothing calls it because the conservative assemblies already impose that condition structurally; routing through the resolver would be redundant on those paths and double-counting on the FEM one. The resolver is not unused because it was forgotten — it encodes a condition its consumers already satisfy by construction.

Whether it should exist at all is now an open design question rather than a wiring bug.

## What a fix owes

1. **Do not add coefficient-reading to the conservative assemblies.** Confirm first, per path, whether the natural BC already is $J\cdot n = 0$; on `divergence_*` and on the FEM weak form it is.
2. **A named refusal for a provider-valued `alpha`/`beta`** on both paths, replacing a silent no-flux wall and a bare `TypeError`. This is the smallest change with real user impact.
3. **A general Robin on the grid paths**, if wanted, must pass `boundary_conditions` into the boundary handlers — a signature change, and only worth it for $\alpha \ne v_n$ or $g \ne 0$.
4. **Document which paths impose the wall structurally.** Two do, neither says so, and that absence is what produced this issue. A test that asserts *mass conservation at a drifted wall* is the check that would have said so — an external oracle, and the one this area was missing.
5. **#1977** — bring the FEM solvers, and the 10 other undeclared solver classes, inside the capability gate.

## The methodological finding, which is the durable part

**"No BC-type dispatch at the boundary" does not imply "imposes $\partial_n m = 0$."** A conservative flux form imposes the correct condition *without naming it*, and so does a particle reflection. I read the absence of a named branch as the absence of the condition, and then measured `ROBIN(alpha=3.2) == no_flux` byte-identically and read that as evidence of a defect.

It was the opposite. With `u = -3.2x`, coupling 1, $\sigma = 0.3$, the resolver emits `ROBIN(alpha=+3.2, beta=-0.045)` — **the same numbers the test used**. That case *is* the physical wall the default scheme already imposes, so byte-identity there is the correct answer. Only $\alpha = 999$ is a genuinely different wall, and only that case shows a real gap.

A census keyed on declarations is structurally blind to correctness that is structural rather than declared. That is the same shape as #1977's finding one level up: the solvers with the most capability declare the least.

---

<details>
<summary>Original text and the two corrections (superseded, kept for the record)</summary>

Filed 2026-08-16 claiming "no FP solver accepts ROBIN — the library's BC support is inverted relative to the mathematics", on a census of `_SUPPORTED_BC_TYPES` across `alg/numerical/{fp,hjb}_solvers/`.

**First correction, same day.** `FPFEMSolver` implements Robin and reads the coefficients. The census used `pkgutil.iter_modules`, which does not recurse, so `alg/numerical/fem/` was never reached; and both FEM solvers declare no `_SUPPORTED_BC_TYPES` at all, so they would have been invisible anyway. That second fact became **#1977**.

**Second correction, 2026-08-17.** The remaining claim — "every grid FP solver imposes $\partial_n m = 0$ unconditionally" — is false for both divergence-form schemes, including the default, as measured above. The prescribed fix would have broken them and the FEM path.

Both errors have one shape: **asserting the extent of a defect from an incomplete sweep, and reading a structural property off a declaration.** Recorded rather than deleted because the correction is the useful part.

</details>

---

## [ADDED 2026-08-17] The rest of the FP wall lanes, measured

The commit that rewrote this issue said the GFDM and network findings were recorded here. They were not — `grep` for either in the body returned 0 hits. Measured now, in a worktree at `7e161ef1` with `PYTHONPATH` exported (`mfgarchon.__file__` verified inside the worktree on every probe).

### `FPGFDMSolver` conflates `NO_FLUX` with homogeneous `NEUMANN`

`_SUPPORTED_BC_TYPES = {NO_FLUX, NEUMANN}` (`fp_gfdm.py:93`). Same problem, $\sigma = 0.3$, wall-normal drift $3.2$, $N = 41$ collocation points, $N_t = 40$:

| input | result |
|:---|:---|
| `no_flux_bc(1)` vs `neumann_bc(0.0, 1)` | **byte-identical**, `max\|diff\| = 0.000e+00` |
| `neumann_bc(+5.0)`, `neumann_bc(-9.0)` | `NotImplementedError: declares BCType.NEUMANN but honours only the homogeneous ...` |
| `dirichlet_bc(7.0)` | `NotImplementedError: does not support boundary condition type(s) ['DIRICHLET']` |

The gate (#1456) works, and the inhomogeneous refusal is loud. What is wrong is narrower: for FP with nonzero wall-normal drift, $J\cdot n = 0$ and $\partial_n m = 0$ are **different conditions** — that distinction is the whole subject of this issue — and this solver returns the same array for both. It declares two types and implements one wall.

Consistent with that, `self._boundary_type = resolved_bc_type` at `fp_gfdm.py:209` is the only occurrence of that attribute in the package: **assigned once, read nowhere.**

Which wall is it? Measured $\partial_n m = +13.99$ at the high wall, so **not** $\partial_n m = 0$ — a scheme imposing that returns exactly `0.0000`, as `gradient_upwind` does in the table above.

**Mass cannot settle the other half, and I am not going to claim it does.** The drifted run reports $-82.2\%$, but the solver itself warns that `upwind_scheme='none'` leaves the flux divergence unstabilised and that it *diverges under refinement* (#1752), so the interior is a confound. Two controls:

```
drift = 0, sigma = 0     mass   +0.000%   d_n m = -8.0e-18
drift = 0, sigma = 0.3   mass   +0.105%   d_n m = +8.8e-04
drift = 3.2, tight blob  ValueError from the mass-fabrication gate at t_idx=1
```

Zero drift is essentially conservative, so the $-82\%$ is not attributable to the wall on this evidence. **Open: which wall the GFDM path imposes is undetermined by mass, and needs #1752 stabilised first or a flux-based probe that does not route through the interior.**

### The network family is a second, parallel BC framework

`FPNetworkSolver` states it outright at `fp_network.py:104`: the network BC model is `NetworkMFGComponents.boundary_nodes`, a node-level Dirichlet pin applied via `NetworkMFGProblem.apply_boundary_conditions`, **"NOT the continuum `BCType`/segments framework"** — so the inherited `_validate_bc_support`, which keys on `BoundaryConditions.segments`, is a *structural no-op* for every network problem.

It has its own capability gate with its own attribute:

- `FPNetworkSolver._honors_node_bc = False` — it ignores `boundary_nodes` and unconditionally renormalises total mass each step, so it cannot honour a node BC.
- `NetworkHJBSolver` raises when `geometry.has_explicit_boundary_conditions()` and `_honors_node_bc` is False (`hjb_network.py:120`, #1468/#1471).
- `NetworkPolicyIterationHJBSolver` sets it True and applies the pin at `hjb_network.py:369`.

This is the concrete answer to *is there another mechanism elsewhere*: **yes, and it is disjoint.** Every census in #1976 and #1980 keys on `BCType`/segments and therefore reports nothing about it. Neither gate knows the other exists.

### 14 separate `apply_boundary_conditions` entry points

| subpackage | count |
|:---|---:|
| `utils` | 5 |
| `alg` | 5 |
| `geometry` | 3 |
| `extensions` | 1 |

Full list: `alg/base_solver.py:380`, `alg/numerical/fp_solvers/fp_particle_bc.py:187`, `alg/numerical/hjb_solvers/hjb_semi_lagrangian.py:2891`, `alg/numerical/hjb_solvers/hjb_sl_characteristics.py:174` and `:289`, `extensions/topology.py:558`, `geometry/boundary/protocols.py:761`, `geometry/implicit/hyperrectangle.py:270`, `geometry/implicit/implicit_domain.py:329`, `utils/acceleration/jax_utils.py:312`, `utils/numerical/particle/boundary.py:40` and `:116`, `utils/particle_utils.py:143` and `:206`.

### The particle helpers are forked, and the fork has not diverged yet

`utils/particle_utils.py` and `utils/numerical/particle/boundary.py` each define `apply_boundary_conditions_numpy` and `apply_boundary_conditions_gpu`. Not a re-export — two independent bodies.

Run on identical inputs across `{periodic, no_flux, dirichlet, reflecting, garbage}` x `{interior, outside both, far outside, exactly on wall, empty, degenerate domain}` = 30 cases: **they agree everywhere**, and both raise `ValueError` on the two unsupported type strings.

So this is a null result, and it is reported as one. The fork is live and **unpinned** — nothing asserts the two agree, which is the case that owes a pinning test before it diverges silently.

One defect did fall out: on a zero-width domain (`xmin == xmax`), both return `[nan nan]` for `periodic` without a word, while the `no_flux` branch of the same function guards with `if Lx > 1e-14`. Inconsistent guarding inside one function.

**Method note.** The 30-case comparison first reported a disagreement on `periodic` + degenerate domain. That was my comparator: `np.allclose` is False for `nan` vs `nan`. `array_equal(..., equal_nan=True)` shows them identical. Recorded because a difference-detector that fires on its own convention is the same class of instrument error this issue was filed under.

derrring:
Refs #1975, #1977, #1979. No production code. **Total rewrite** — the first version of this file got the physics backwards and I self-blocked it; the independent review then confirmed that independently and found three more things.

## What was wrong

It read "no BC-type dispatch at the boundary" as "imposes $\partial_n m = 0$", measured `ROBIN(alpha=3.2) == no_flux` byte-identically, and called that a defect.

The reflecting wall is $J\cdot n = 0$, **not** $\partial_n m = 0$. The conservative schemes impose it **structurally**, by zeroing the total face flux, with no branch naming it. And with `u = -3.2x`, coupling 1, $\sigma = 0.3$, `FPResolver` emits `ROBIN(alpha=+3.2, beta=-0.045)` — *the same numbers the test used*. So that case **is** the physical wall the default scheme already imposes, and byte-identity there is the correct answer.

## The load-bearing test is now an external oracle

$\sigma = 0.3$, wall-normal drift 3.2, $n_x = 81$, 200 steps:

| scheme | mass drift | $\partial_n m$ at the wall |
|:---|---:|---:|
| `gradient_centered` | −78.05 % | +3.7 |
| `gradient_upwind` | −75.47 % | +2.1 |
| `divergence_centered` | **−0.0000 %** | +1962 |
| `divergence_upwind` **(default)** | **−0.0000 %** | +1138 |

Mass conserved to machine precision at a drifted wall **with a nonzero normal gradient there** is only possible under $J\cdot n = 0$. Positive control, refining against $\partial_n m = (v/D)m$: `0.7377 / 0.8491 / 0.9184 / 0.9574` at $n_x = 201/401/801/1601$, monotone to 1 at first order.

Mass conservation is a law of the equation, computed here without reference to any scheme's internals. **Its absence is exactly what let the wrong reading stand.** A census keyed on declarations cannot answer a question about behaviour.

## The population is now discovered independently of what it audits

`walk_packages` + `issubclass(BaseNumericalSolver)` — a predicate that does not depend on the declaration being measured. The old version enumerated by reading `_SUPPORTED_BC_TYPES` inside two non-recursed directories, so it could not report its own blind spot: it missed **10** solvers, and its completeness guard compared the discovered set against a frozen copy of *the same discovery*.

**Ungated solvers are recorded as a population, not as an absence**: ~~11 of 22~~ **10 of 21** concrete solvers declare nothing, `FPFEMSolver` among them (#1977) [CORRECTED 2026-08-17 — computed: `len(_population())` is 21, of which 11 declare].

~~Two further classes — `HJBHowardSolver`, `ImplicitHeatSolver` — apply BC segments without being solver subclasses at all.~~ [CORRECTED 2026-08-17] **Three** classes are named rather than discovered, and only one of them does what that sentence claimed:

| class | what its own source does with a BC segment |
|---|---|
| `HJBHowardSolver` | reads `seg.bc_type` and branches on `DIRICHLET` — `hjb_howard.py:363` |
| `ImplicitHeatSolver` | **nothing.** Its only mention is inside an f-string in a diagnostic — `implicit_heat.py:293` |
| `BoundaryHandler` | full BC vocabulary; **omitted from this list entirely** in the previous revision |

The pattern is the one that produced the sentence: `grep` for BC vocabulary found the words in all three and I wrote "apply" for all three. Vocabulary present ≠ condition imposed. They are **named** rather than discovered because a population predicate is itself a scope claim, and the honest form says what it does not cover — but naming them is not licence to assert, without tracing, what they do.

## Mutations

`sha256` verified identical after each. Old file → new.

**X6, re-measured at `8be0d4dd`** — the divergence wall swapped for the gradient wall in `_BOUNDARY_HANDLERS`, verified live by asserting the dispatched handler's `__name__` at import rather than by diffing the file:

| file shape | X6 |
|---|---|
| with the three behavioural tests | **2 failed**, 27 passed |
| without them (the rev4 strip) | **0 failed**, 24 passed |

That is what the fourth review meant by *the deletion cost more than the prose*, and it is why they are back.

| mutation | old | new |
|---|---|---|
| X1 the gate stops raising | **16 P — uncaught** | 1 F / 24 P |
| X3 behaviour-preserving rename of the gate method | **1 F — false positive** | 25 P |
| X4 `FPFDMSolver` declares ROBIN, assembly unchanged | would have read as the fix | 2 F / 23 P |
| X5 a solver one subpackage deeper | **16 P — invisible** | 1 F / 24 P |
| X6 wall flux balances diffusion only | n/a | 2 F / 23 P |

X1 and X3 replace an `inspect.getsource` string match the review showed was wrong in **both** directions. X4 is the naive fix #1975 warns against — declaration without implementation — now caught by a coupled assertion. X6 is the physics oracle firing on $J\cdot n = 0$ degenerating to $\partial_n m = 0$.

**Honest limit on X6**: it kills `divergence_upwind` and the family-separation control but **not** `divergence_centered`, so that row is the less pinned of the two and I am not claiming otherwise.

The byte-identity assertion is re-scoped to `alpha=999`, the only case that is a genuinely different wall.

## Review findings, disposition

| finding | disposition |
|:---|:---|
| census materially incomplete (blocker 1) | fixed — `walk_packages` + subclass predicate |
| population guard cannot notice its own omission (blocker 2) | fixed — independent predicate; ungated set pinned as a population |
| "every grid FP solver imposes $\partial_n m = 0$" false for the default (blocker 3) | fixed — claim deleted, replaced by the conservation oracle |
| `inspect.getsource` wrong in both directions | fixed — behavioural `pytest.raises` |
| declaration-only state passes the assembly check | fixed — coupled assertion (X4) |
| mutation table mixed two file revisions | fixed — re-run whole against one revision |
| the PR adds two files, not one | acknowledged; the changelog fragment is rewritten too |
| "imports many solver modules at collection time" — false | correct, it was my error in the brief; collection is light |

Also carried into #1975 (rewritten), #1977 and #1979 (new).

## Gate

`./scripts/local_ci.sh` → **6234 passed, 12 skipped, 21 xfailed, GATE GREEN**, 156 s at `8be0d4dd`.
(+5 restored parametrisations − 1 deleted gate-premise test = +4 from the 6230 above.)
Run in a dedicated worktree with `PYTHONPATH` exported — the gate strips cwd, so without it it measures the shared checkout.

### DIFF
--- a/changelog.d/1975-fp-boundary-coefficient-census.added.md
+++ b/changelog.d/1975-fp-boundary-coefficient-census.added.md
@@ -1,19 +1,35 @@
 A test file recording which solvers declare their BC support and that the FDM boundary assembly
 reads none of a ROBIN segment's coefficients (#1975, #1977, #1979).
 
-Assertions only. Four revisions put the reasoning, the history and the physics in the file as
-prose, and four independent reviews found a false statement in that prose each time -- five wrong
-numbers, one wrong sign, and a correction of a reviewer that was itself wrong. Nothing is now
-stated that the file does not compute; the narrative lives in #1975.
+Assertions only. Five revisions put the reasoning, the history and the physics in the file as
+prose, and five independent reviews found a false statement in that prose each time -- six wrong
+numbers, one wrong sign, a correction of a reviewer that was itself wrong, and an argument for
+keeping a filter that measurement showed did nothing. No NUMBER is now stated that the file does
+not compute; the traced source references it carries instead each name a file and a line.
 
 The population is discovered by `walk_packages` + `issubclass` + `inspect.isabstract`, keyed on
-class identity, with the root package imported explicitly and cross-module re-exports collapsed by
-the defining-module filter. Verified against injected escapes: a solver in `alg/__init__.py`, one
-named `Baseline*`, a factory-built class, an alias sorting alphabetically earlier, and the deletion
-of a dead alias.
+class identity, with the root package imported explicitly and the canonical name pinned to
+`cls.__name__` so an alphabetically earlier alias cannot displace it. There is no defining-module
+filter: it gained nothing (21 classes with it, 21 without) and hid one real cross-module alias,
+`HJBNetworkSolver` at `network_solvers/__init__.py:19`, from the very test written to catch that
+case. Verified against injected escapes: a solver in `alg/__init__.py`, one named `Baseline*`, an
+alias sorting alphabetically earlier, a cross-module alias, and the deletion of a dead alias.
+
+A factory-built class whose `__module__` points at another scanned module is NOT caught -- that
+escape is real as a mechanism and has no instance in this tree (measured: 0 classes defined in a
+scanned module and never bound there).
 
 The gate reads `supported_bc_types`; this file reads `_SUPPORTED_BC_TYPES`. Those coincide only
-while every property forwards, so that premise is now itself asserted: all 11 declaring solvers
-have a property whose body is exactly `return self._SUPPORTED_BC_TYPES`, and the 11 that declare
-nothing have no property at all. Widening the live gate through the property while leaving the
-private attribute alone previously left the file green.
+while every property forwards, and that premise is NOT asserted. A revision asserted it by
+comparing `inspect.getsource(...).splitlines()[-1]`; review measured that open in three shapes and
+falsely closed on a fourth, so it is deleted rather than repaired and the hole is stated in the
+file: a solver can widen its live support through the property alone and every assertion stays
+green. Of 21 concrete solvers, 11 declare and 10 do not.
+
+Two disclosed scope limits. The population predicate is `BaseNumericalSolver`, while
+`_validate_bc_support` and `honors_inhomogeneous_neumann` live on `BaseMFGSolver`: 4 concrete
+classes (`PrimalDualMFGSolver`, `SinkhornMFGSolver`, `VariationalMFGSolver`,
+`WassersteinMFGSolver`) are in the gate's reach and outside the census. None mentions a boundary
+condition anywhere, so none hides a live defect. And the behavioural oracle is 1D with
+`interface_velocity=None`, so the conservative wall's multi-axis branch and its interface-velocity
+branch are exercised by nothing here.
--- a/changelog.d/1975-fp-boundary-coefficient-census.added.md
+++ b/changelog.d/1975-fp-boundary-coefficient-census.added.md
@@ -1,0 +1,7 @@
+A census pin over every solver's `_SUPPORTED_BC_TYPES`, plus two assertions on the gap it
+records (#1975). The Fokker-Planck reflecting wall is Robin in `m`; no FP solver accepts
+`BCType.ROBIN`, the only solver that does is on the HJB side where the condition is Neumann, and
+`FPResolver` emits exactly the type every consumer refuses. Measured: a mixed `ROBIN(alpha=999)`
+wall assembles byte-identically to no-flux, so `(alpha, beta, g)` is read by nothing on the FP
+side and the gate refusing ROBIN is honest rather than an oversight. Both gap assertions pin
+neither side, so they go green when the gap closes from either end.
--- a/changelog.d/1975-fp-boundary-coefficient-census.added.md
+++ b/changelog.d/1975-fp-boundary-coefficient-census.added.md
@@ -1,7 +1,16 @@
-A census pin over every solver's `_SUPPORTED_BC_TYPES`, plus two assertions on the gap it
-records (#1975). The Fokker-Planck reflecting wall is Robin in `m`; no FP solver accepts
-`BCType.ROBIN`, the only solver that does is on the HJB side where the condition is Neumann, and
-`FPResolver` emits exactly the type every consumer refuses. Measured: a mixed `ROBIN(alpha=999)`
-wall assembles byte-identically to no-flux, so `(alpha, beta, g)` is read by nothing on the FP
-side and the gate refusing ROBIN is honest rather than an oversight. Both gap assertions pin
-neither side, so they go green when the gap closes from either end.
+A test file recording which wall each FP path actually imposes, and which solvers sit outside the
+BC capability gate (#1975, #1977, #1979).
+
+Its load-bearing test is an external oracle the area was missing: **mass conservation at a wall
+with wall-normal drift.** The conservative schemes -- `divergence_upwind`, the default, and
+`divergence_centered` -- impose `J.n = 0` structurally by zeroing the total face flux, conserving
+mass to machine precision with `d_n m` nonzero at the wall; the `gradient_*` family imposes
+`d_n m = 0` and loses 75-78% (non-conservative by design, #1075). Neither behaviour was asserted
+anywhere, and that absence is what let this file's own first version read the right answer as a
+defect.
+
+The census now discovers its population with `walk_packages` + `issubclass`, a predicate
+independent of the declaration it audits, and records the **ungated** solvers as a population
+rather than as an absence: 11 of 22 concrete solvers declare no `_SUPPORTED_BC_TYPES`, including
+`FPFEMSolver`, the one FP solver implementing a general Robin. Two further classes apply BCs
+without being solver subclasses at all and are named rather than discovered.
--- a/changelog.d/1975-fp-boundary-coefficient-census.added.md
+++ b/changelog.d/1975-fp-boundary-coefficient-census.added.md
@@ -1,16 +1,14 @@
-A test file recording which wall each FP path actually imposes, and which solvers sit outside the
-BC capability gate (#1975, #1977, #1979).
+A test file recording which solvers sit outside the BC capability gate, and that the FDM boundary
+assembly reads none of a ROBIN segment's coefficients (#1975, #1977, #1979).
 
-Its load-bearing test is an external oracle the area was missing: **mass conservation at a wall
-with wall-normal drift.** The conservative schemes -- `divergence_upwind`, the default, and
-`divergence_centered` -- impose `J.n = 0` structurally by zeroing the total face flux, conserving
-mass to machine precision with `d_n m` nonzero at the wall; the `gradient_*` family imposes
-`d_n m = 0` and loses 75-78% (non-conservative by design, #1075). Neither behaviour was asserted
-anywhere, and that absence is what let this file's own first version read the right answer as a
-defect.
+The population is discovered by `walk_packages` + `issubclass` + `inspect.isabstract`, keyed on
+class identity: a predicate independent of the declaration it audits. Two same-module alias pairs
+(`NetworkFPSolver = FPNetworkSolver`, `NetworkHJBSolver = HJBNetworkSolver`) are collapsed and
+pinned separately, the root package is imported explicitly because `walk_packages` never yields it,
+and the `startswith("Base")` name heuristic is gone -- it excluded any concrete solver whose name
+began with "Base" and bought nothing `isabstract` did not already cover. Verified against four
+injected escapes: a solver in `alg/__init__.py`, one named `Baseline*`, a factory-built class, and
+the deletion of a dead alias.
 
-The census now discovers its population with `walk_packages` + `issubclass`, a predicate
-independent of the declaration it audits, and records the **ungated** solvers as a population
-rather than as an absence: 11 of 22 concrete solvers declare no `_SUPPORTED_BC_TYPES`, including
-`FPFEMSolver`, the one FP solver implementing a general Robin. Two further classes apply BCs
-without being solver subclasses at all and are named rather than discovered.
+No count is restated in prose. Three independent attempts at this file produced three different
+counts, each written down as a fact; the frozen sets are the measurement.
--- a/changelog.d/1975-fp-boundary-coefficient-census.added.md
+++ b/changelog.d/1975-fp-boundary-coefficient-census.added.md
@@ -1,35 +1,32 @@
 A test file recording which solvers declare their BC support and that the FDM boundary assembly
 reads none of a ROBIN segment's coefficients (#1975, #1977, #1979).
 
-Assertions only. Five revisions put the reasoning, the history and the physics in the file as
-prose, and five independent reviews found a false statement in that prose each time -- six wrong
-numbers, one wrong sign, a correction of a reviewer that was itself wrong, and an argument for
-keeping a filter that measurement showed did nothing. No NUMBER is now stated that the file does
-not compute; the traced source references it carries instead each name a file and a line.
+Assertions, with the reasoning in #1975. Six revisions put the reasoning in the file as prose and
+six independent reviews found a false statement in it each time -- including in the sentences that
+announced the file's own discipline, which were false in the same commit that added them. Those
+sentences are gone; what the file can compute it asserts, and what it cannot it cites at a line.
 
 The population is discovered by `walk_packages` + `issubclass` + `inspect.isabstract`, keyed on
 class identity, with the root package imported explicitly and the canonical name pinned to
 `cls.__name__` so an alphabetically earlier alias cannot displace it. There is no defining-module
-filter: it gained nothing (21 classes with it, 21 without) and hid one real cross-module alias,
-`HJBNetworkSolver` at `network_solvers/__init__.py:19`, from the very test written to catch that
-case. Verified against injected escapes: a solver in `alg/__init__.py`, one named `Baseline*`, an
-alias sorting alphabetically earlier, a cross-module alias, and the deletion of a dead alias.
-
-A factory-built class whose `__module__` points at another scanned module is NOT caught -- that
-escape is real as a mechanism and has no instance in this tree (measured: 0 classes defined in a
-scanned module and never bound there).
+filter. Removing it gained no classes and admitted one name -- the cross-module alias
+`HJBNetworkSolver` at `network_solvers/__init__.py:19`, which the test written for that case could
+not see -- and it also closed the factory escape, since a `type()`-built class over an `ABCMeta`
+base gets `__module__ == 'abc'` and was rejected in every scanned module while the filter stood.
+Verified against injected escapes: a solver in `alg/__init__.py`, one named `Baseline*`, an alias
+sorting alphabetically earlier, a cross-module alias, a factory-built class with a retargeted
+`__module__`, and the deletion of a dead alias.
 
 The gate reads `supported_bc_types`; this file reads `_SUPPORTED_BC_TYPES`. Those coincide only
 while every property forwards, and that premise is NOT asserted. A revision asserted it by
 comparing `inspect.getsource(...).splitlines()[-1]`; review measured that open in three shapes and
 falsely closed on a fourth, so it is deleted rather than repaired and the hole is stated in the
 file: a solver can widen its live support through the property alone and every assertion stays
-green. Of 21 concrete solvers, 11 declare and 10 do not.
+green.
 
-Two disclosed scope limits. The population predicate is `BaseNumericalSolver`, while
-`_validate_bc_support` and `honors_inhomogeneous_neumann` live on `BaseMFGSolver`: 4 concrete
-classes (`PrimalDualMFGSolver`, `SinkhornMFGSolver`, `VariationalMFGSolver`,
-`WassersteinMFGSolver`) are in the gate's reach and outside the census. None mentions a boundary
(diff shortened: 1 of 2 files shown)


derrring:
## Author self-block: the census is incomplete

Found while doing #1975's item 4 (the false Robin declarations), after this PR was opened and after the review was dispatched.

**`FPFEMSolver` implements Robin and reads `(alpha, beta, g)`.** `fem/fp_fem_solver.py:92` overrides `_robin_operator_terms` → `assemble_robin_terms` (`fem/bc_adapter.py:255`), assembling $D(\alpha/\beta)\int_{\partial\Omega}\phi_i\phi_j$ and load $D(1/\beta)\int_{\partial\Omega}g\,\phi_i$:

```
sum|A|(alpha=999) / sum|A|(alpha=3.2)  = 312.187500     999/3.2 = 312.187500
sum|A|(beta=-.045) / sum|A|(beta=-.090) =   2.000000     expect 2.0
sum|rhs|(g=7) = 28.0        sum|rhs|(g=0) = 0.0
```

So this PR's docstring claim — *"no FP solver in this library accepts `BCType.ROBIN`"* — is false, and `_CENSUS` is missing two solvers.

**Two independent reasons the discovery function misses them.** `pkgutil.iter_modules` over `fp_solvers/` and `hjb_solvers/` does not recurse and never reaches `alg/numerical/fem/`; and both FEM solvers declare **no** `_SUPPORTED_BC_TYPES` at all, so they would be invisible even from the right directory. That second fact is its own defect and is now **#1977** — the two solvers with the most BC capability are the two the `_validate_bc_support` gate does not cover.

**And the census's own guard cannot catch this.** `test_the_solver_population_has_not_changed` compares the discovered population against a frozen copy of *the same discovery*. A census whose population is defined by the predicate it audits cannot detect an omission in that predicate. The fix is to enumerate by **subclass** (`BaseFPSolver` / `BaseHJBSolver` concrete subclasses), which does not depend on the class declaring anything, and to assert separately that every concrete solver declares a support set.

What survives unchanged: the byte-identity measurement on the FDM assembly (`alpha=999` ≡ no-flux), the six mutations, and the "does this defend the defect" design. The **grid** FP paths really do ignore the coefficients — it is the scope claim that was wrong, not the measurement.

#1975 is now marked `[PARTIALLY-SUPERSEDED]` with the correction and an amended scorecard.

I will rewrite the census on a subclass-based population before asking for merge. The review in flight has been sent this correction and asked to adjudicate it independently rather than take it from me, plus two things I have **not** measured: whether an `FPFEMSolver` solve with $\alpha = v_n$ actually conserves mass at a wall with wall-normal drift (`assemble_robin_terms` puts the advective half of the flux out of scope, citing #1237, and $J\cdot n = 0$ needs both halves), and how each path behaves on a provider-valued `alpha` — which on the FEM path raises a bare `TypeError` from `float()` rather than a named refusal.

derrring:
## Correction: this PR's framing came from a census that has since been rebuilt

Two things in the body above are now wrong, both inherited from `capability_census`'s lane 2 rather than from anything in this file.

**1. "Every grid FP solver here imposes `d_n m = 0` unconditionally" — false, and this PR's own docstring says it.** The conservative schemes impose `J.n = 0` structurally by zeroing the total face flux. Measured, $\sigma=0.3$, wall-normal drift 3.2:

| scheme | mass drift | $\partial_n m$ at the wall |
|:---|---:|---:|
| `gradient_centered` | −78.05 % | +3.7 |
| `gradient_upwind` | −75.47 % | +2.1 |
| `divergence_centered` | **−0.0000 %** | +1962 |
| `divergence_upwind` **(default)** | **−0.0000 %** | +1138 |

**2. Mass conservation is neither sufficient nor necessary, so it must never be a verdict.** Not sufficient: streamline diffusion conserves to $10^{-12}$ while the wall ratio collapses $0.967 \to 0.414$. Not necessary: `FPSLJacobianSolver` is the Lagrangian form, non-conservative **by construction**, deprecated for adjoint inconsistency rather than for mass — the census called it `CONTROL_FAILED`, a legitimate scheme reported as broken.

The lane-2 instrument has since been rebuilt (#1980, `c1d996e3`) after **eight** defects, four of which surfaced only on re-running it against a known answer. The sharpest: it passed a potential to the three solvers whose declared `_drift_convention` is `VELOCITY`, so the wall-normal drift vanished at the wall the mass reached — the discriminating property was absent and a verdict printed anyway.

## What in THIS PR survives, and it is most of it

The assertions here do not depend on either wrong claim:

- **the byte-identity measurement stands.** `_BOUNDARY_HANDLERS` handlers are not passed `boundary_conditions` at all — the parameter is absent from the signature — so no configuration can make them read a coefficient. 768 combinations, zero non-identical cases. The `alpha=999` scoping is correct and is the case that shows a real gap.
- **the population predicate stands** (`walk_packages` + `issubclass`, independent of the declaration audited), and so does the ungated-population row: 11 of 22 concrete solvers declare nothing.
- **the behavioural gate assertion stands**, replacing an `inspect.getsource` match that was wrong in both directions.
- **the coupled assertion stands** — a declaration without an implementation still fails.

## What needs amending here

The module docstring's sentence *"Every grid FP solver here imposes `d_n m = 0` unconditionally"* is false and should read: the **`gradient_*`** family imposes `d_n m = 0` and is non-conservative by design (#1075); the **`divergence_*`** family imposes `J.n = 0` structurally. I will amend it rather than leave a PR that ships a false physics claim in a test docstring — which is the same defect class this whole campaign has been fixing in `fp_fdm.py:802` and `robin_bc`.

Related: #1975 (rewritten twice), #1977, #1979, #1980.

derrring:
## Correction to my own previous comment

I wrote that the module docstring still carries *"Every grid FP solver here imposes `d_n m = 0` unconditionally"* and said I would amend it. **It does not.** That sentence was in the version this file replaced; the rewrite (`8a97b00e`) already states the correct physics in its opening paragraph:

> **The reflecting wall is `J.n = 0`, not `d_n m = 0`.** They coincide only when the drift is tangential at the wall. The conservative schemes impose the first **structurally** — by zeroing the total face flux, advective plus diffusive — with no BC-type branch naming it; the `gradient_*` family imposes the second and leaks.

`grep -c "Every grid FP solver here imposes"` on the branch head returns **0**.

I described the previous revision while looking at a comment thread rather than the file — the same "read the label, not the artifact" move this campaign keeps recording. No amendment is needed and none will be made; the review in flight has been told so it does not chase an absent defect.

The rest of that comment stands: the two claims in the **PR body** were inherited from a census since rebuilt (#1980, `c1d996e3`, eight instrument defects), and the body's lane-2 framing is superseded by the comment above it. The assertions in this file — byte-identity, the population predicate, the behavioural gate assertion, the coupled declaration/implementation assertion — are unaffected by any of it.

Issue #1977, opened 2026-08-16
The BC capability gate does not cover the FEM solvers: FPFEMSolver and HJBFEMSolver declare no _SUPPORTED_BC_TYPES


derrring:
Split out of #1975, where it was the reason a census went wrong.

## Measured

```
FPFEMSolver      _SUPPORTED_BC_TYPES = NONE -- ungated
HJBFEMSolver     _SUPPORTED_BC_TYPES = NONE -- ungated
```

Every other solver in `alg/numerical/{fp,hjb}_solvers/` declares one and is checked by `_validate_bc_support` at construction (`alg/base_solver.py:314/324`). The two FEM solvers declare nothing, so nothing is checked.

## Why it matters more than it looks

The gate exists because of #1456: a solver handed a BC type it does not implement would *silently apply a different wall*. The two solvers outside it are the two with the **most** BC capability in the library — `FPFEMSolver` is the only FP solver that implements Robin at all (`_robin_operator_terms` → `assemble_robin_terms`, `fem/bc_adapter.py:255`), and it reads every coefficient:

```
sum|A|(alpha=999) / sum|A|(alpha=3.2)  = 312.187500     999/3.2 = 312.187500
sum|A|(beta=-.045) / sum|A|(beta=-.090) =   2.000000     expect 2.0
sum|rhs|(g=7) = 28.0        sum|rhs|(g=0) = 0.0
```

So the gap is not "an unimplemented solver escaped a check". It is that the solver whose BC support most needs *stating* is the one that states nothing — a user cannot find out what an FEM solve honours except by reading the assembly, and no test can assert it.

## Concretely undeclared today

`assemble_robin_terms`'s own docstring records limits that nothing enforces at construction:

- **`beta == 0`** (pure Dirichlet written as Robin) — documented to fail loud, but from inside the assembly.
- **Callable / `BCValueProvider` `g`** — documented as unimplemented.
- **Provider-valued `alpha`** — not documented, and fails with a bare builtin error rather than a named one: `TypeError: float() argument must be a string or a real number, not 'AdjointConsistentProvider'`.
- **The advective half of a flux wall** — "the advection-flux boundary coupling for an inhomogeneous Robin OUTFLOW is a distinct total-flux BC and is out of scope (Issue #1237)". A caller has no way to learn this from the solver.

None of these is a declaration a caller can query; all four are discoverable only by hitting them.

## Also: this defeated a census

[...]

PR #1978, opened 2026-08-16
docs(fp): correct three declarations that said FP-FDM supports Robin (#1975)

derrring:
Refs #1975 item 4. Documentation and one diagnostic string. No behaviour change.

## The three

1. **`fp_fdm.py:802`** — the deprecation docstring for `solve_fp_system_1d` advertised the unified nD solver as *"Full BC support (no_flux, neumann, robin, periodic, dirichlet)"*.
2. **`fp_fdm_time_stepping.py:1314`** — the fail-loud diagnostic for a legacy BC told users to *"Use `mfgarchon.geometry.boundary` (periodic_bc / dirichlet_bc / **robin_bc** / no_flux_bc) with the modern BoundaryConditions instead"*. The same file cannot assemble what it recommends. This one is the worst of the three: it routed a user from a **loud** failure to a **silent** one.
3. **`robin_bc`'s own docstring** said nothing about which solvers honour it.

## The fact

`_BOUNDARY_HANDLERS` (`fp_fdm_time_stepping.py:110`) is keyed on the *advection scheme*, not on BC type, and all four entries are `add_boundary_no_flux_entries_*`. The call site is `elif (is_no_flux or not is_uniform) and is_boundary`, so a mixed BC at a boundary point takes the no-flux handler whatever its segments say. One assembled step:

```
mixed NO_FLUX               max|m - m_noflux| = 0.000e+00   IDENTICAL
mixed ROBIN(a=3.2,b=-.045)  max|m - m_noflux| = 0.000e+00   IDENTICAL
mixed ROBIN(a=999,b=-.045)  max|m - m_noflux| = 0.000e+00   IDENTICAL
```

## Why these are corrections, not removals

Robin **is** honoured elsewhere in the library — which I got wrong in #1975's first draft and corrected there:

- **`FPFEMSolver` / `HJBFEMSolver`** assemble it in weak form via `assemble_robin_terms` and read every coefficient. Measured: `sum|A|(alpha=999)/sum|A|(alpha=3.2) = 312.187500` against `999/3.2 = 312.187500`; `beta` exactly inverse; `g` into the load.
- **`HJBGFDMSolver`** takes the adjoint-consistent `Robin(0, 1)` case.

So `robin_bc`'s docstring now says which solvers honour it, which refuse it, **and** the limits: constant `g` only, `beta == 0` fails loud, and a provider-valued `alpha` raises a bare `TypeError` from `float()`. The FDM diagnostic now points at `FPFEMSolver` instead of just dropping `robin_bc` from the list.

Related: **#1977** — both FEM solvers declare no `_SUPPORTED_BC_TYPES`, so none of those limits is checked at construction.

## Documentation discipline

Struck rather than rewritten where the text shipped: `~~Full BC support (...)~~` followed by `[CORRECTED 2026-08-16, #1975]` and the measurement. A claim that shipped and was wrong is not silently replaced.

## Gate

`./scripts/local_ci.sh` → **6206 passed, 12 skipped, 21 xfailed, GATE GREEN**, 195 s.

Run inside a dedicated worktree with `PYTHONPATH` exported. Worth recording: the gate invokes `PYTHONSAFEPATH=1 "$PY" -P -m pytest`, and both flags strip cwd, so **running `local_ci.sh` from inside a worktree without exporting `PYTHONPATH` silently measures the shared checkout instead.** Measured:

```
cwd=WT, no PYTHONPATH, plain                     -> WT/mfgarchon        (cwd wins)
cwd=WT, no PYTHONPATH, -P + PYTHONSAFEPATH=1     -> shared checkout     (editable finder wins)
cwd=WT, PYTHONPATH=WT, -P + PYTHONSAFEPATH=1     -> WT/mfgarchon        (PYTHONPATH wins)
cwd=/tmp, PYTHONPATH=WT, -P + PYTHONSAFEPATH=1   -> WT/mfgarchon        (PYTHONPATH wins)
```

A plain `pytest` and the gate can therefore resolve to *different trees* in the same shell.

### DIFF
--- a/changelog.d/1975-false-robin-declarations.fixed.md
+++ b/changelog.d/1975-false-robin-declarations.fixed.md
@@ -1,7 +1,7 @@
 Three shipped declarations said the FP-FDM path supports Robin. It does not: the boundary handlers
 are not passed `boundary_conditions` at all, so a ROBIN segment assembles byte-identically to
 no-flux (measured at alpha=3.2 and alpha=999 across 768 configurations). Corrected
-`solve_fp_system_1d`'s deprecation docstring, the legacy-BC diagnostic that recommended `robin_bc`
+`FPFDMSolver._solve_fp_1d`'s deprecation docstring, the legacy-BC diagnostic that recommended `robin_bc`
 on a path that cannot assemble it, and `robin_bc`'s own docstring.
 
 `robin_bc` now also says what none of the three said and what matters most in practice: **you do
@@ -9,5 +9,5 @@ not need a ROBIN segment for a reflecting wall.** The conservative schemes -- `d
 (the default) and `divergence_centered` -- impose `J.n = 0` structurally by zeroing the total face
 flux, conserving mass to machine precision at a wall with wall-normal drift; the `gradient_*`
 family imposes `d_n m = 0` instead and loses 75-78% (non-conservative by design, #1075). Handing
-the FEM path `alpha = v_n` explicitly double-counts the same term. Reach for `robin_bc` only for a
+adding a Robin segment on top of it destroys that wall: `A_robin` contributes a residual outflux `J.n = D*(alpha/beta)*m`, so the implied wall is `D d_n m = (v_n - D*alpha/beta) m`. Reach for `robin_bc` only for a
 wall that is not the reflecting one. (#1975)
--- a/changelog.d/1975-false-robin-declarations.fixed.md
+++ b/changelog.d/1975-false-robin-declarations.fixed.md
@@ -1,13 +1,22 @@
 Three shipped declarations said the FP-FDM path supports Robin. It does not: the boundary handlers
 are not passed `boundary_conditions` at all, so a ROBIN segment assembles byte-identically to
-no-flux (measured at alpha=3.2 and alpha=999 across 768 configurations). Corrected
-`FPFDMSolver._solve_fp_1d`'s deprecation docstring, the legacy-BC diagnostic that recommended `robin_bc`
-on a path that cannot assemble it, and `robin_bc`'s own docstring.
+no-flux. Corrected `FPFDMSolver._solve_fp_1d`'s deprecation docstring, the legacy-BC diagnostic
+that recommended `robin_bc` on a path that cannot assemble it, and `robin_bc`'s own docstring.
 
-`robin_bc` now also says what none of the three said and what matters most in practice: **you do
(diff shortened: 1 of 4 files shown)


derrring:
## Second commit — the replacement text was still built on #1975's wrong premise

The three false declarations this PR identified were correctly identified. What the first commit put in their place assumed a reflecting wall with wall-normal drift is unavailable on the grid paths. **It is available, and it is the default.**

`add_boundary_no_flux_entries_conservative` (`fp_fdm_bc.py:222`, aliased to `divergence_upwind`) zeroes the **total** face flux — advective plus diffusive — which is $J\cdot n = 0$. Measured, $\sigma = 0.3$, wall-normal drift 3.2, $n_x = 81$, 200 steps:

| scheme | mass drift | $\partial_n m$ at the wall |
|:---|---:|---:|
| `gradient_centered` | −78.05 % | +3.7 |
| `gradient_upwind` | −75.47 % | +2.1 |
| `divergence_centered` | **−0.0000 %** | +1962 |
| `divergence_upwind` **(default)** | **−0.0000 %** | +1138 |

Mass conserved to machine precision at a drifted wall **with $\partial_n m$ nonzero there** is only possible under $J\cdot n = 0$. Positive control, refinement against $\partial_n m = (v/D)\,m$: ratio `0.7377 / 0.8491 / 0.9184 / 0.9574` at $n_x = 201/401/801/1601$, monotone to 1 at first order. `gradient_upwind` returns exactly `0.0000` — that family does impose $\partial_n m = 0$ and is non-conservative by design (#1075).

So `robin_bc`'s docstring now leads with what a user actually needs: **use `no_flux_bc` on a conservative scheme for a reflecting wall**, and reach for `robin_bc` only when $\alpha \ne v_n$ or $g \ne 0$. Handing the FEM path $\alpha = v_n$ explicitly **double-counts** it — the weak form's natural BC is already $J\cdot n = 0$, so the implied wall becomes $D\,\partial_n m = 2 v_n m$ and mass drifts −79.5%.

The underlying point, and the reason this PR needed a second pass: **two of the library's FP paths impose the correct wall structurally rather than by a named branch** — the conservative FDM schemes, and `FPParticleSolver` via Skorokhod reflection — and until now neither said so anywhere a reader could find. That absence is exactly what produced #1975's wrong premise, twice.

Also recorded in the docstring: a provider-valued `alpha` is **silently accepted** on the FDM path and produces a no-flux wall with no diagnostic — now **#1979**.

`./scripts/local_ci.sh` → **6206 passed, 12 skipped, 21 xfailed, GATE GREEN**, 142 s.

PR #1980, opened 2026-08-16
feat(scripts): a capability census — what classes declare, and what FP walls actually do

derrring:
Refs #1975, #1977. Adds `scripts/capability_census.py` and its ratchet `tests/unit/test_capability_census.py`. No changes to library code.

> **[REWRITTEN 2026-08-17]** — the previous body documented a second lane deleted two commits ago (`e3b3123c`) and carried five numbers measured against that vanished version: `16 of 26`, `19 solvers`, two `1 F / 21 P` mutation rows against a 22-test file, and a gate line of `6228 passed / 171 s`. Everything below is measured at `1728fb2b`.

## Why this exists

The 2026-08-13 design census had four lanes, and **every one looks for reality falling short of a claim.** Nobody counted the other direction, and the other direction is what made #1975 wrong twice: a solver that implements a capability and declares nothing is invisible to a sweep that reads declarations. `FPFEMSolver` implements a general Robin wall and says so nowhere a grep can find.

## What it measures

Population by `pkgutil.walk_packages` + `issubclass` against four named roots, then own-vs-inherited by walking the MRO. The predicate does not depend on the property being audited.

```
=== declares nothing of its own ===
  applicator    4 of   5  (80%)
  backend       4 of   4  (100%)
  geometry     15 of  15  (100%)
  solver       15 of  25  (60%)
```

Declarations that arrive by inheritance are now grouped by `(owner, value)` rather than counted:

```
  honors_inhomogeneous_neumann: own 6, inherited 19
       18  <- BaseMFGSolver            = True
        1  <- FPSLSolver               = False
  _SUPPORTED_BC_TYPES: own 10, inherited 1
        1  <- FPSLSolver               = frozenset({PERIODIC, NEUMANN, NO_FLUX})
  discretization_type: own 2, inherited 24
       20  <- BaseNumericalSolver      = FDM
        4  <- three applicator bases   = <property>
```

**18 solvers claim to honour an inhomogeneous Neumann flux by a default nobody chose.** The count alone said 19 and called all of them permissive; the 19th inherits `False` — a deliberate refusal, the opposite of the claim. Two of the three rows were not permissive defaults at all. That is why the section groups now.

## Mutations

Every mutant's identity asserted at import before reading the result; every file restored and `sha256`-verified after.

| mutation | result |
|---|---|
| M1 — `BaseMFGSolver.honors_inhomogeneous_neumann` `True → False` | **1 F / 8 P** |
| M3 — `FPFEMSolver` starts declaring `_SUPPORTED_BC_TYPES` | **1 F / 8 P** |
| MH — `FPFDMSolver`'s own `honors_inhomogeneous_neumann` `False → True` | **1 F / 8 P** (was 0 F before this round) |
| ML — a gated solver widens `_SUPPORTED_BC_TYPES` to include ROBIN | **0 F / 9 P** — uncovered here, see below |

## The one stated limit, and its cover

ML is not caught by this file and is not meant to be: pinning per-solver declared values here would duplicate #1976's frozen copy, which is the divergence this census exists to find. Verified that the cover is real — the same ROBIN mutation against #1976 at `8be0d4dd`: **2 failed, 27 passed**, one of them `test_each_gated_solver_declares_exactly_this[FPFDMSolver]`. **#1976 must merge before this PR.**

The previous revision claimed #1976 would pin "the declared VALUES". It pins one of the three: `_SUPPORTED_BC_TYPES`. `honors_inhomogeneous_neumann`'s own values were pinned by neither file — MH passed 8/8 here and 29/29 there — so this round pins them here, where there is nothing to duplicate. `discretization_type`'s two owners declare a `property` object whose `str()` carries a memory address; not pinnable this way, and stated as uncovered.

## What this cannot say

- **A name still decides part of the population.** `startswith(("Base", "_"))` drops 4 concrete classes under a root — `BaseGraphApplicator`, `BaseMeshfreeApplicator`, `BaseStructuredApplicator`, `BaseUnstructuredApplicator` — all with `inspect.isabstract` False, and all four *own* `discretization_type` (three appear as the `<-` owner in the output; the fourth has no surviving subclass). The script now prints them rather than dropping them silently. A name sweep deciding a population is the reason the second lane was deleted; this is the residue.
- **A namespace-style subpackage is invisible.** `iter_modules` never yields a directory without `__init__.py`. Current count of such directories: **0** (372 modules on disk, 372 walked).
- **Instance-level declarations are invisible.** `self._SUPPORTED_BC_TYPES = ...` in `__init__` would be honoured by the runtime gate and unseen here. No class does this today.
- **The JSON artifact is not reproducible** — `cls_id` and `<property object at 0x…>` differ between runs, and frozenset ordering moves with `PYTHONHASHSEED`. The fields the tests read are stable; the artifact is not.

## Gate

`./scripts/local_ci.sh` → **6215 passed, 12 skipped, 21 xfailed, GATE GREEN**, 164 s at `1728fb2b`.
Run in a dedicated worktree with `PYTHONPATH` exported — the gate strips cwd, so without it it measures the shared checkout.

### DIFF
--- a/changelog.d/1975-capability-census.added.md
+++ b/changelog.d/1975-capability-census.added.md
@@ -1,23 +1,20 @@
-`scripts/capability_census.py` and its ratchet: what each class **declares**, and the wall-ratio
-**sequence** each FP path produces.
+`scripts/capability_census.py` and its ratchet: what each class **declares** about the boundary
+conditions it accepts.
 
-Lane 1 discovers its population with `walk_packages` + `issubclass` + `inspect.isabstract`, keyed
-on class identity -- a predicate independent of the declaration it audits -- so "declares nothing"
-is a recorded row rather than an absence, and own declarations are separated from inherited ones by
-walking the MRO.
+The 2026-08-13 design census had four lanes and every one looks for reality falling *short* of a
+claim; it found 77 over-claims. Nobody counted the other direction, and that is what made #1975
+wrong -- `FPFEMSolver` implements a general Robin wall and declares nothing, so a census keyed on
+`_SUPPORTED_BC_TYPES` reported the capability absent.
 
-**Lane 2 renders no verdict.** An earlier version classified each path from the ratio's trend
-across three resolutions; independent review showed the rule cannot do that. `FPFVMSolver` reads
-0.392 / 0.649 / 1.106 and keeps going to 12.699 at nx=1281, so a "within tolerance of 1" clause
-fires on a value the sequence merely transits, and `FPSLSolver` reads 0.998 at nx=201 then 1.926
-and 3.500. Three resolutions cannot separate approach from transit, and neither can six. The
-sequence is reported and the reading is left to a person.
+The population comes from `walk_packages` + `issubclass`, a predicate independent of the
+declaration audited, so "declares nothing" is a recorded row rather than an absence; own
+declarations are separated from inherited ones by walking the MRO. `honors_inhomogeneous_neumann`
+defaults to `True` on `BaseMFGSolver`, so 20 solvers claim to honour an inhomogeneous flux by a
+default nobody chose. Three classes apply BC segments without being subclasses of any root and are
+named rather than discovered; that list is stated to be incomplete.
 
-Mass drift is reported beside it as a **form property**: neither sufficient (streamline diffusion
-conserves to 1e-12 while the ratio collapses) nor necessary (`FPSLJacobianSolver` is the Lagrangian
-form, non-conservative by construction, deprecated for adjoint inconsistency rather than for mass).
-
-Also folded: the empty deprecated subclass `FPSLAdjointSolver(FPSLSolver)`, via its own
-`_deprecation_meta["alias_for"]` -- class-keying collapses `X = Y` but not `class X(Y): pass`, and
-two identical rows read as two independent confirmations. The clip exemption is keyed on the
-declared `kde_boundary_smoothing` flag rather than a substring of the class name. (#1975, #1977)
+A second lane measuring which wall each FP path imposes was removed. Its findings are recorded in
+#1975. It needed a discrimination rule for the LIMIT behaviour of a numerical scheme -- four
+attempts failed to state one, and 41% of a 32-mutation sweep survived the ratchet built over it,
+including the pins for three of the defects that ratchet claimed to have fixed. A ratchet over an
+unstated rule pins nothing.
--- a/changelog.d/1975-capability-census.added.md
+++ b/changelog.d/1975-capability-census.added.md
@@ -1,12 +1,10 @@
 `scripts/capability_census.py` and its ratchet `tests/unit/test_capability_census.py`: what each
-class **declares**, and what the FP paths actually **do** at a wall.
+class **declares**, and which wall each FP path actually **imposes**.
 
 Both lanes answer a question the 2026-08-13 design census could not. All four of its lanes look for
 reality falling *short* of a claim, and it found 77 over-claims; nobody counted the other
-direction. The other direction is what made #1975 wrong twice — `FPFEMSolver` implements a general
-Robin wall and declares nothing, and the default FDM scheme imposes `J.n = 0` structurally with no
-branch naming it. Over-claiming makes a user's code fail loudly; under-claiming makes a maintainer
-"fix" something that works.
+direction, and the other direction made #1975 wrong twice. Over-claiming makes a user's code fail
+loudly; under-claiming makes a maintainer "fix" something that works.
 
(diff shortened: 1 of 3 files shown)


derrring:
## The instrument had eight defects. Every one printed a confident verdict.

The lane-2 numbers in the body above are **wrong** and are superseded by `c1d996e3`. Four defects were found by independently measuring the five paths this census could not construct; four more only by re-running it against a path whose answer was already known.

| # | defect | found by |
|---|---|---|
| 1 | assumed the mass functional | independent measurement |
| 2 | verdict was sign-blind | independent measurement |
| 3 | could not tell a stability failure from a wall | independent measurement |
| 4 | double-counted a module-level alias | independent measurement |
| 5 | assumed which wall was the outflow wall | re-run, 1st |
| 6 | clip gate false-positived on a particle method | re-run, 1st |
| **7** | **assumed the drift convention** | re-run, 2nd |
| 8 | thresholded a quantity meaningful only as a limit | re-run, 3rd |

### Defect 7 is the one to keep

`_drift_convention` is a **declared class attribute** reading `VELOCITY` on `FPFDMSolver`, `FPFVMSolver` and `FPGFDMSolver`, whose second positional argument is `drift_field`, not a potential. This census passed all of them `u = -3.2x`, consumed as the velocity field $\alpha(x) = -3.2x$ — which **vanishes at $x=0$**, the wall the mass then piled against.

Those rows had **no wall-normal drift at the measured wall**. The discriminating property was absent and a verdict printed anyway.

Lane 1 of this same script exists to argue that capability must be read from declarations rather than assumed. Lane 2 did not read the declaration sitting on the class.

(Related: `FPParticleSolver`'s parameter is named `drift_field` while its declared convention is `VALUE_FUNCTION`. Name and declaration disagree; the harness now follows the declaration and pins the disagreement.)

### The verdict is no longer mass conservation

It is the wall ratio $\partial_n m / ((v_n/D)\,m_{\text{wall}})$ and its **trend** across $N_x = 41/81/161$. Mass drift is reported beside it as a **form property**:

- **not sufficient** — streamline diffusion conserves to $10^{-12}$ while the ratio collapses $0.967 \to 0.414$;
- **not necessary** — `FPSLJacobianSolver` is the Lagrangian form $m^{n+1}(x) = m^n(x-\alpha\,dt)e^{-dt\,\mathrm{div}\alpha}$, non-conservative by construction with an $O(h)$ error that halves under refinement, and deprecated for **adjoint inconsistency**, not for mass. The old instrument called it `CONTROL_FAILED` — a legitimate scheme reported as broken.

### Current rows

```
FPFDMSolver         0.346 -> 0.519 -> 0.685   IMPOSES_J_DOT_N
FPFVMSolver         0.392 -> 0.649 -> 1.106   IMPOSES_J_DOT_N
FPSLSolver         -0.007 -> 0.041 -> 0.465   IMPOSES_J_DOT_N
FPSLAdjointSolver  -0.007 -> 0.041 -> 0.465   IMPOSES_J_DOT_N
FPSLJacobianSolver  0.046 -> 0.026 -> 0.010   IMPOSES_ZERO_GRADIENT   (mass +2.43%)
FPParticleSolver    witness does not transfer to a binned density -- recorded, not judged
5 NOT_MEASURED
```

`FPFDMSolver`'s trend reproduces the independent direct measurement of the raw assembly (`0.4592 / 0.6438 / 0.7933` at the same resolutions) — the known answer this rebuild was validated against, and the check the original never had.

`FPParticleSolver` is **recorded, not judged**: the ratio is defined from $\partial_n m$ on a represented density, and a binned/KDE reconstruction is a different object, with the reflection living in the paths. A witness that does not apply is not evidence either way — reading its output as a verdict is the same error as reading `NOT_MEASURED` as a pass.

### What the five `NOT_MEASURED` rows are, from independent measurement

Recipes in #1975. `FPFEMSolver` and `MeshlessGalerkinFPSolver` impose $J\cdot n = 0$ as the **natural BC** of the weak form, with no boundary term assembled at all. `FPNetworkSolver` conserves structurally (960/960 configurations, max $|drift| = 8.8\times10^{-12}$%) and has no boundary in this sense — the drift is a CTMC generator on edges with no off-graph target. `WeakFormFPSolver` is not independently constructible. **`FPGFDMSolver` imposes neither condition**: its `_boundary_type` is resolved, gate-validated, stored at `fp_gfdm.py:209` and never read again, ratio flat at 0.077, −80.2% with 99.6% of the loss accounted for by the wall flux.

### Gate

`./scripts/local_ci.sh` → **6230 passed, 12 skipped, 21 xfailed, GATE GREEN**, 128 s. Ratchet runs in 9.5 s.

The PR body is now stale in its lane-2 half; this comment supersedes it.

Issue #1992, opened 2026-08-17
Coupling interface vs the MFG taxonomy: three entry points with no slot


derrring:
Systematic read of the MFG literature (Carmona–Delarue I & II, Cardaliaguet's lecture notes, Gomes–Pimentel–Voskanyan, CDLL *Master Equation*, Bardi–Cirant, Bensoussan–Frehse–Yam, Sun–Yong), mapped against this library's coupling interface. Every formula cited below was read from a rendered page, not a PDF text layer — the text layers of these books drop leading minus signs, overlines, and primes, which inverts exactly the statements a taxonomy turns on.

## The interface, as measured

| slot | sees `m` as | `p`-dependent? | inside the α-minimisation? |
|---|---|---|---|
| `H(x_idx, m_at_x, derivs)` | **scalar at x** | yes | yes |
| `source_term_hjb(x, m_t, v_t, t)` | full array | via `v_t`, numerically | **no** |
| `source_term_fp(x, m_t, v_t, t)` | full array | same | n/a |
| `nonlocal_operator @ v_t` | not at all — acts on `v` | linear | no |

The operative constraint is **not** "can `m` enter". It is that the Hamiltonian is `p`-dependent but sees `m` only pointwise, while the source is nonlocal in `m` but sits **outside the minimisation** — so it is invisible to the monotone/upwind numerical Hamiltonian and to the policy-iteration linearisation. Anything needing both at once falls between the two.

## What the design covers

Most of the mainline, and one result is worth stating plainly because it went the other way from my expectation: **congestion is handled correctly, on both sides.** `CongestionHamiltonian` implements `H = H_kin(p)/c(m) + f(m)` with analytic `∂H/∂m = -c'(m)·H_kin/c(m)² + f'(m)`, matching Gomes–Pimentel–Voskanyan p. 116. The FP side then routes it correctly: `CongestionHamiltonian` is a sibling of `SeparableHamiltonian`, not a subclass, so `isinstance(H_class, SeparableHamiltonian)` is False in `resolve_fp_drift_kwargs`, the velocity path is taken, and `compute_fp_velocity_field` evaluates `H.optimal_control(x, m_faces, p, t)` on face-averaged density `m_{i+1/2}`. That is `m·D_pH(x,∇u,m)` with the correct staggering — the transport term `div(m^{1-α}Du)` of Gomes (7.2), not `div(m∇u)`.

[...]

PR #1995, opened 2026-08-18
docs: the hamiltonian example is not congestion

derrring:
Refs #1992. One docstring, three wrong things in it — and it is the example a reader copies.

## It is not congestion

`BaseMFGProblem.hamiltonian` labelled this "Quadratic Hamiltonian with congestion":

```python
return 0.5 * np.sum(p_arr**2) + 0.1 * m
```

Congestion means moving is costlier where the crowd is dense, so the density scales the **momentum** term:

$$H(x,p,m)=\frac{m^{\alpha}}{\gamma}\left(\frac{|p|}{m^{\alpha}}\right)^{\gamma},\qquad \text{i.e. } \frac{|p|^2}{2m^{\alpha}} \text{ at } \gamma=2$$

— Gomes, Pimentel & Voskanyan, *Regularity Theory for Mean-Field Game Systems*, p. 116, with the stated motivation *"agents face difficulties in moving at high speed in high-density areas"*. Every congestion model in that corpus is pointwise in `m` and puts it in the momentum term.

In the example `m` sits in a term with no `p` in it at all. Decomposed against Gomes (1.18), it is `H_0(x,p) - F(x,m)` — a **separable Hamiltonian with a local coupling**, the simplest form in the taxonomy.

The library already implements the real thing in `CongestionHamiltonian`, with an analytic `dH/dm`. The docstring now points there.

## The sign was the non-uniqueness one

Under the convention pairing `H` with `-u_t + H = 0` (Gomes (1.18), Cardaliaguet (4.13), CDLL (1.7), Bardi–Cirant (1.1) all use it), `F(x,m) = -0.1m` gives

$$\int\big(F(x,m_1)-F(x,m_2)\big)(m_1-m_2)\,dx = -0.1\int(m_1-m_2)^2 < 0$$

— crowd-**attracting**, violating Lasry–Lions monotonicity, in the regime where the known counterexamples to uniqueness live. Now `-0.1 * m`.

Worth saying what this is *not*: a non-monotone sign is perfectly legitimate in a manufactured verification instance, and the GFDM paper's own MMS takes it deliberately (`chapters/appendix.tex`: *"its congestion enters with the sign opposite… so it is not Lasry–Lions monotone… Neither affects the measured order, which compares a computed solution against an imposed exact one"*). The objection is to it being the **default example**, not to the sign as such.

## `m` is a scalar, and the docstring never said so

Added: `m` is the density **at** `x`, not the measure. A coupling needing the whole density — a convolution `(k*m)(x)`, the standard nonlocal form and the one the mainline MFG literature actually uses — cannot be written in this slot at all. It has to go through `source_term_hjb(x, m_t, v_t, t)`, which receives the full array but is additive and therefore sits **outside** the optimal-control minimisation.

That last distinction is the one that decides which MFG forms this interface can express, so it belongs where a reader implementing a Hamiltonian will see it.

## Gate

Docstring only, no behaviour change. `tests/unit/test_core`: 698 passed, 3 xfailed. Full gate green via the pre-push hook.

## One qualification on this PR's own framing

Review measured that `MFGProblem` — the library's main concrete problem class — does not implement `hamiltonian` at all (`hasattr(MFGProblem, "hamiltonian")` is `False`; it uses `H(x_idx, m_at_x, derivs, t_idx)` plus `hamiltonian_class`). `MFGProblemProtocol` is a `typing.Protocol` with a `...` body, `@runtime_checkable`, which checks attribute presence only — so nothing enforces it, and the protocol's own class docstring asserting `isinstance(problem, MFGProblemProtocol)` is false.

So "the example a reader copies" is softer than written. Real implementers of the 4-argument form exist — `examples/advanced/capacity_constrained_mfg/problem.py:149` and four test files — but the library's flagship class is not one of them. That does not weaken the correctness argument: a false statement in the norm is worth fixing whether or not traffic reaches it, and fixing the instances first would have meant correcting them against a specification that was itself wrong. It does mean the priority claim was overstated. Tracked on #1996, which also carries the same congestion mislabel in that example.

### DIFF
--- a/changelog.d/1992-congestion-docstring.fixed.md
+++ b/changelog.d/1992-congestion-docstring.fixed.md
@@ -1,0 +1,7 @@
+`BaseMFGProblem.hamiltonian`'s docstring no longer calls `0.5*|p|^2 + 0.1*m` a congestion
+Hamiltonian. It is a separable Hamiltonian with a local coupling: congestion means the density
+scales the momentum term, `H = (m^alpha/gamma)(|p|/m^alpha)^gamma` (Gomes-Pimentel-Voskanyan
+p. 116), whereas there `m` sits in a term with no `p` in it. The example now points to
+`CongestionHamiltonian` for the real form, flips the sign to `-0.1*m` so the default is
+Lasry-Lions monotone rather than crowd-attracting, and states that `m` is the scalar density at
+`x` and not the measure, so a nonlocal coupling has to go through `source_term_hjb`.
--- a/changelog.d/1992-congestion-docstring.fixed.md
+++ b/changelog.d/1992-congestion-docstring.fixed.md
@@ -1,5 +1,5 @@
-`BaseMFGProblem.hamiltonian`'s docstring no longer calls `0.5*|p|^2 + 0.1*m` a congestion
-Hamiltonian. It is a separable Hamiltonian with a local coupling: congestion means the density
+`MFGProblemProtocol.hamiltonian`'s docstring no longer calls `0.5*|p|^2 + 0.1*m` a
+congestion Hamiltonian. It is a separable Hamiltonian with a local coupling: congestion means the density
 scales the momentum term, `H = (m^alpha/gamma)(|p|/m^alpha)^gamma` (Gomes-Pimentel-Voskanyan
 p. 116), whereas there `m` sits in a term with no `p` in it. The example now points to
 `CongestionHamiltonian` for the real form, flips the sign to `-0.1*m` so the default is
--- a/mfgarchon/core/base_problem.py
+++ b/mfgarchon/core/base_problem.py
@@ -120,10 +120,21 @@ class MFGProblemProtocol(Protocol):
             Hamiltonian value H(x, m, p, t)
 
         Note:
-            ``m`` is the density AT ``x``, a scalar -- not the measure. A coupling that needs the
-            whole density, such as a convolution ``(k * m)(x)``, cannot be written here; use
-            ``source_term_hjb(x, m_t, v_t, t)``, which receives the full array but is additive and
-            therefore sits outside the optimal-control minimisation.
+            On the grid and mesh interface ``m`` is the density AT ``x``, a scalar -- not the
+            measure. (``NetworkMFGProblem.hamiltonian`` takes a different, 5-argument form and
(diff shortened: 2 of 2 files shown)


derrring:
Review note: the two mathematical claims here were verified independently before this PR existed, by a separate agent reading the sources directly rather than by me. The congestion definition is quoted from Gomes–Pimentel–Voskanyan p. 116 read from a **rendered page** (`pdftoppm`), not the text layer — that PDF encodes `=` as `D` and `+` as `C` and drops every Greek and relational glyph, so its text layer is unusable for exactly this kind of claim. The monotonicity integral was computed independently and reaches the same sign.

The convention-dependence is stated in the PR body and is the one thing a reviewer should press on: the classification (separable + local, not congestion) is convention-independent, the monotonicity verdict is not. It rests on the library forming its HJB residual as `-u_t + H = 0`, which is what Gomes (1.18), Cardaliaguet (4.13), CDLL (1.7) and Bardi–Cirant (1.1) all use.

derrring:
## The convention-dependent claim, now checked against the code

The PR body flags that the classification (separable + local, not congestion) is convention-independent while the **monotonicity verdict is not** — it depends on how this library forms its HJB residual. The source-reading agent could not settle that, having never read the code. Settled now, from the artifact:

`hjb_fdm.py:665-670`
```python
rhs = H_values
if source_term is not None:
    rhs = rhs - source_term
return U_next - self.dt * rhs
```

and `base_hjb.py:746` — `F(u) = ... - S = 0`, `Phi_U -= source_term`.

So `U^n = U^{n+1} - dt*(H - S)`, i.e. `-(U^{n+1} - U^n)/dt + H = S`, i.e.

```
-u_t + H = S
```

That is exactly the convention Gomes (1.18), Cardaliaguet (4.13), CDLL (1.7) and Bardi–Cirant (1.1) all use. Writing the example as `-u_t + H_0(p) = F(x,m)` therefore gives `F(x,m) = -0.1m` for the original `+0.1*m`, hence

```
∫ (F(x,m₁) − F(x,m₂))(m₁ − m₂) dx = −0.1 ∫ (m₁ − m₂)² dx < 0
```

— crowd-attracting, Lasry–Lions monotonicity violated. The sign flip in this PR is correct under the library's own convention and not only under the literature's.

Issue #1904, opened 2026-08-12
The no-flux/Neumann ghost is cell-centred on a node-centred grid: the wall Laplacian converges to HALF the true value


derrring:
`pad_array_with_ghosts` writes a **cell-centred** mirror, `ghost = u[0]`, onto a **node-centred** grid — `TensorProductGrid.get_grid_spacing()` is `span/(N-1)` and the coordinates include both endpoints, so the boundary sits *on* node 0. The wall equation is therefore inconsistent: it converges to a different PDE than the interior.

Found while reviewing #1902, which deleted a post-solve boundary overwrite on the grounds that "the residual owns the no-flux condition". It does not.

## Measured

`u = cos(2*pi*x)` has `u_x = 0` at both walls exactly and `u_xx(0) = -4*pi^2`. Ratio of the discrete wall Laplacian to the true value:

| Nx | `lap[0] / u_xx(0)` | interior node, same run |
|--:|--:|--:|
| 21 | 0.495901 | 0.991802 |
| 41 | 0.498973 | 0.997946 |
| 81 | 0.499743 | 0.999486 |
| 161 | 0.499936 | 0.999871 |
| 321 | **0.499984** | 0.999968 |

The interior converges to 1. **The wall converges to 1/2.** It is not a slow rate — it is the wrong limit.

The ghost, probed directly on `u = [1,2,3,4,5]`:

| BC | ghost_lo | ghost_hi | node-centred would give |
|:--|--:|--:|:--|
| `no_flux` | 1.000 | 5.000 | 2.000 / 4.000 |
| `neumann(0)` | 1.000 | 5.000 | 2.000 / 4.000 |

`ghost = u[0]` states `(u[0] - ghost)/h = 0`, i.e. a zero derivative half a cell *outside* the domain. On a node-centred grid the second-order statement is the reflection `ghost = u[1]`, which gives `lap[0] = 2*(u[1] - u[0])/h^2`.

## The remedy restores consistency

Same fixture, both ghosts:

| Nx | cell-centred (current) | node-centred `u[-1] = u[1]` |
|--:|--:|--:|
| 21 | 0.495901 | 0.991802 |
| 321 | 0.499984 | **0.999968** |

Independent measurement during the #1902 review, by manufactured solution `u(t,x) = (1+t)cos(2*pi*x)` (linear in `t`, so backward Euler is exact and only the spatial error is measured):

| Nx | ghost | L2 error | order | node 0 error | order |
[...]

PR #1997, opened 2026-08-18
chore: a branch classifier that prints evidence and does not delete

derrring:
**Redesigned after review: this classifies and prints evidence. It does not delete.**

43 local branches had accumulated, 30 of them dead. Cleaning them required *deriving* each branch's fate, and that derivation would have to be redone every time — so I automated it, including the deletion. Adversarial review found five blockers, three of which share one root: every signal it uses can be wrong in a way that costs unmerged work.

## The one that settles it

`git apply --check --reverse` targets the **working tree**, not `main`. Nothing pins the checkout, the cwd, or a clean tree. Same script, same repo, same minute:

```
$ bash scripts/prune_local_branches.sh          # from repo root
dry run: 0 branch(es) would be deleted.

$ cd scripts && bash ./prune_local_branches.sh  # one cd
... 9 total
```

Those nine are live branches — including both `#1904` branches with the issue open, and `diag/fd-jacobian-instrumentation-1116`, which the original PR body named as the branch its "MERGED only" rule existed to protect. `git apply` skips patch paths outside the cwd, so from `scripts/` **zero hunks are checked and it exits 0 with no output at all**. The verdict also moves with uncommitted edits and with which branch is checked out — and my own validation run was on `chore/prune-local-branches`, not `main`.

Two more of the same shape:

- **Tier 1 compared names, never shas.** 8 branch names in this repo have been the head ref of more than one merged PR. A reused name plus new commits is deleted.
- **Tier 3's `[0-9]{3,4}` is unanchored.** `perf/gfdm-10000-particles` → `1000` → merged → delete. In a numerical codebase, grid sizes and particle counts in branch names are ordinary. And tier 3 fires *only* on branches tier 2 reported as not-in-main, so every false positive there destroys unmerged work by construction.

## The fix is to remove the automation, not to patch the signals

Patching each would leave a tool whose correctness rests on three subtle invariants, run rarely, whose failure mode is silent destruction. `--delete` is gone; any argument is refused. The report ends by naming its own three limits, so a reader knows what the labels do not prove:

```
0 branch(es) look disposable. This tool does not delete them, deliberately:
  - the content check is run against the WORKING TREE, not against main, ...
  - the merged-PR check matches head-ref NAMES, and names are reused after a merge;
  - the number extracted from a branch name may be a grid size, not an issue.
Read the branch, then: git branch -D <name>   (record the sha first).
```

The value that survives is real: it does the 1121-item merged-PR fetch, the per-branch content test and the worktree check in one command, so the *evidence* is assembled without being re-derived. The judgement stays where it was.

## Also fixed

- Private `mktemp -d` workspace with a trap, replacing fixed world-writable names in `/tmp` — the delete list among them, which a concurrent run could truncate mid-sweep, and which a symlink could redirect onto a real file.
- An explicit numeric check on the merged-PR total. `gh ... -q .total_count` **exits 0 with empty output** if the key is missing, which made `$((TOTAL / 2))` zero and silently disabled the very guard this tool was written around.
- AGENTS.md's claim that `git status` "cannot see worktrees at all" was wrong. Verified both directions: an ignored path (`.gitignore:77`, `.claude/worktrees/`) shows *nothing*; a non-ignored path shows `?? wt/`. What is never reported is the size. Corrected here and in the cross-project guidance that carried the same sentence.

## Note on the review

The reviewer flagged that seven branches vanished from the working tree mid-review. That was me, cleaning branches manually in the same repo while the review ran — my coordination error, not theirs. Their B1 measurement was captured against the earlier state and is recorded as such; I reproduced the cwd behaviour independently afterwards.

Gate green: 6271 passed, 12 skipped, 23 xfailed.


derrring:
Converting to draft. Independent review returned MERGE-BLOCKED with four blockers, and the heaviest
one needs a different predicate rather than a patch.

## What works and should survive any rewrite

The reverse-apply predicate **does** solve the squash-merge problem. Verified in a throwaway repo
across 13 branches: a squash-merged branch absent from the merged-PR name list returns
`DELETE (content in main)` on the content check alone — which `git rev-list --count` and
`git branch --merged` structurally cannot do. A branch landed then reverted correctly returns `keep`.
That is this PR's real contribution.

## Blockers

1. **From any subdirectory, every branch is labelled `DELETE`** — 13 of 13, including branches whose
   commits exist nowhere else. `git apply` silently skips patch paths outside cwd, so with no paths
   left it exits 0. Control: the identical invocation from the repo root on the identical state
   returns 5 DELETE / 8 keep, so the sweep is not silently empty.

   **This is not the defect the shipped caveat describes.** Lines 78–79 blame the working tree.
   Tested the fix that sentence points at — a pristine `origin/main` index via `GIT_INDEX_FILE` +
   `read-tree` + `--cached`: correct from root, **still exit 0 from a subdirectory**. The path skip
   happens before any content comparison, so a reader who fixes the documented cause still has the
   13/13 failure. The word "cwd" appears in neither the script, nor AGENTS.md, nor the fragment.

2. **A mode-only change gives a false `DELETE`.** `100644 → 100755`, same blob, branch not in main
   → `DELETE (content in main)`. `git apply --check --reverse` does not verify file mode. This is
   the unrecoverable direction, from the repo root, on a clean tree.

3. **A dirty working tree gives a false `DELETE` on the only copy of the work.** One uncommitted
   edit matching the branch is enough; the "evidence" is something `git checkout -- .` erases.

4. **The file's own docstring asserts the behaviour the latest commit forbids.** Line 2 still says
   "and delete the ones that did"; lines 10–12 document a `--delete` flag and a recovery manifest
   that no longer exist; `STAMP`/`MANIFEST` are orphans. `shellcheck` fails SC2034 on `MANIFEST`
   (not in any repo gate, so latent). Line 24 also still carries the "`git status` cannot see
   worktrees at all" claim that `1b2f4439` corrected in AGENTS.md — the diff ships both versions.

## And a correction to my own PR body: there is no test

I wrote "plus its test". The diff is three files — `AGENTS.md`, the changelog fragment, and the
script. Nothing under `tests/`. All three of the measured blockers above are first-run failures a
~60-line pytest over `tmp_path` repos would have caught, which is exactly the argument for having
one.

## The rework

The reviewer verified a predicate that is **13/13 correct on the same fixtures** and reads no
working tree, no index and no cwd:

```
git merge-tree --write-tree origin/main <branch>   # compare first line to origin/main^{tree}
```

Right on mode, on binary, on squash, on revert. (Two apparent misses were a fixture artifact —
byte-identical file contents triggering spurious rename detection; `-c merge.renames=false` resolves
both.) That is the direction, and it needs its own test before adoption.

Draft rather than closed: the predicate insight is worth keeping and the rewrite is small.

Separately, and out of scope here: `scripts/check_branch_health.sh:79,123` still recommends
`gh pr list --state merged ... | xargs git push origin --delete`, name-keyed, unguarded, against the
**remote** — the precise hazard this PR was written to eliminate, still live in the same directory.
Worth its own issue.

Issue #1728, opened 2026-07-25
Upwind's spatial order is uncovered: the MMS test cited for it runs at zero drift


derrring:
No test in this repository can fail if `divergence_upwind` stops being first-order in space.

This was found while deleting `test_upwind_first_order_convergence` (#1724). That test could not
measure an order — it compared `result.max_error`, the final Picard residual — so removing it lost
nothing. The deletion's justification, however, named a replacement that does not work either, and
that is the finding.

## The cited coverage runs at zero drift

`TestMMSFokkerPlanck1D::test_sinusoidal_periodic_convergence` (`tests/integration/test_mms_validation.py:221-286`)
measures error against a manufactured solution across a refinement sequence and asserts
`ratios > 1.8`. It solves with `potential_field=U_zero` (`:260`) — no advection. Substituting a
different advection scheme leaves its output bit-identical:

```
divergence_upwind    errors=[0.01243362 0.00640331 0.00323375]  ratios=[1.942 1.980]
divergence_centered  errors=[0.01243362 0.00640331 0.00323375]  ratios=[1.942 1.980]
gradient_upwind      errors=[0.01243362 0.00640331 0.00323375]  ratios=[1.942 1.980]
```

The O(h) signal it measures is spatial — holding `Nt=4000` and refining `Nx` still gives
`[1.955, 1.987]` — but it originates in the diffusion/BC treatment, not in the stencil the ratios
are attributed to.

So the scheme's stated order has never been under test. A refactor that turned `divergence_upwind`
into a centered form would leave the suite green.

## What a real test needs

- A manufactured solution with **non-zero drift**, so the advection term is active.
- A refinement sequence, with the observed order computed rather than a single grid.
- **Mutation-red**: swapping `divergence_upwind` for `divergence_centered` must fail it. That is the
  check the current MMS test fails, and the one that distinguishes measuring a scheme from
  measuring the grid.
- Low enough cell Péclet that centered is not simply unstable — otherwise the test discriminates on
  stability rather than on order.

Found by independent review of #1724; not introduced by it.

Refs #1442, #1701, #1724

PR #2012, opened 2026-08-19
test: delete a "h-convergence" test that compares two unconverged Picard residuals (#1998)

derrring:
> **This body was rewritten at `6f63c22e`+.** The original text was the v1 account, and five review
> rounds falsified four of its claims. It is replaced rather than annotated because leaving both
> would license opposite conclusions with nothing marking which is current — the failure the last
> commit on this branch was written to fix, in the one place I had not checked. What v1 said, and
> why each part is wrong, is in *Corrections* at the bottom.

## What this does

Deletes `test_mesh_refinement_improves_accuracy` from `tests/validation/test_duality_convergence.py`
and replaces it with an account of why it measured nothing. The code diff is one deletion; **the
recorded account is the deliverable.**

## Why the test measured nothing

It compared the final Picard residual of an `Nx=20` solve against an `Nx=40` solve and called the
ratio a convergence rate. Neither solve converges. Measured over the full residual history at
`max_iterations=300`:

```
Nx=20   peak 1.264638e+03 @iter4   best 1.778986e-06 @iter194   @iter29 5.935364e-04
Nx=40   peak 1.243941e+03 @iter4   best 7.116063e-06 @iter166   @iter29 4.237936e-04
```

Both legs run to the iteration cap with `converged=False`. The quantity being compared is *how far
each iteration got in 30 steps*, which depends on the residual's oscillation phase, not on `h`. That
is why the verdict is platform-dependent and why it flipped in opposite directions on darwin and
linux across one commit.

## The provenance

The green-to-red window is `179e55a7..4a50e27b`, and `c98a9c5f` (#1902) is in it — the commit that
deletes the hand-rolled `u[0] = u[1] - g*dx` no-flux enforcement, which is the exact boundary
treatment this test runs through. It moves the coarse leg ~5× on darwin, moves the fine leg *down*
1.31×, and moves the ratio the assertion actually tests from 0.2136 to 1.4005 — a factor 6.56 —
flipping the darwin verdict.

Scope, stated because it was once overstated: **bisected on darwin, inferred on linux.** Only the
endpoints are known on linux. The inference is strong because `c98a9c5f` is the only commit in the
window that touches this computation, but it is an inference.

## What the deletion costs, and what it does not

**The `e < 1000` assertion was a live guard, not a vacuous one.** The residual crosses 1000 by 26%
in the same solve, at iteration 4. It bounded the transient decay rate.

**It was a coarse one.** Tripping it needs the whole decay curve delayed by 24–26 of the 30
iterations, or the per-iteration decay factor to fall from ~1.85 to ≤ 1.0088:

```
Nx=20   delays d with e[29-d] > 1000:  [25]
Nx=40   delays d with e[29-d] > 1000:  [24, 25, 26]
decay factor leaving e[29] exactly 1000:  1.0094 / 1.0088
actual median post-peak decay factor:     1.841 / 1.857
```

A solve that merely slowed down stays green; only one that has stopped converging fires it.

**The transient is not left unguarded at `Nx=41`.** `test_dual_fdm_pair_converges`'s `:94`
assertion (`final_error <= initial_max * 2.0`) reads threshold `1.003133e+03` against final
`1.823556e+02`, a 5.50× margin — and it is **more sensitive**: `:94` fires at a decay factor below
1.0440, `e < 1000` at below 1.0088, so `:94`'s firing region strictly contains it, twenty iterations
sooner.

**What is actually lost is the coarse leg**, `Nx_points=[21]` at `sigma=0.1`, which has no
`:94`-bearing neighbour. As of today's review that cost is *larger* than this PR previously stated —
see Corrections C5.

## Corrections — five rounds, and what each falsified

| v1 claim | status |
|---|---|
| "No commit in the green-to-red window changed the computation … not a regression" | **False.** The two commits I compared for identical digits are both *after* the window. `c98a9c5f` is inside it and moves the numbers. My three earlier bisections probed commits strictly after it and compared pass counts rather than values, so they could not have seen it. |
| Precedent is `test_centered_fdm_higher_order` | **Wrong sibling.** That was removed for a different reason (one grid, a `converged or iterations >= 30` tautology). The precedent for *this* reason is `test_upwind_first_order_convergence`. |
| "`e < 1000` is vacuous" | **False**, peak `1.26e+03`. Then over-corrected: "roughly eight iterations of convergence speed" was wrong by 3× — it is 24–26. |
| "Nothing is lost / strictly stronger than `e < 1000`" | **False.** `max_error` at iteration 29 is entirely the U increment (`err_U 4.238e-04` vs `err_M 4.807e-08`) while `test_fdm_upwind_stable` constrains M. The two assertion sets are **incomparable, not ordered.** |
| "The transient is unguarded" | **False**, `:94` bounds it more sensitively. Corrected in two of three places and left standing in the third — the fix for that is on this branch. |
| C5 (today) "the coarse leg is not uncovered in kind: `fdm_1d` asserts convergence, a 10× drop, mass and positivity" | **False, and it replaced a true statement.** Those four assertions are on the **`fvm_1d`** fixture, which runs `FVM_MUSCL`. `fdm_1d` runs `FDM_UPWIND` and has exactly one consumer, `test_1d_fvm_fdm_agreement`, asserting cross-scheme agreement (`rel_m < 0.07`) and nothing else. So the coarse leg has **no counterpart anywhere** — which is what the account said before this citation overwrote it. |

The habit that found the last one is worth keeping: after changing a claim, grep the diff's own
vocabulary across the file, the changelog fragment **and** the linked issue's title. It found three
instances after I had already "fixed" the claim once.

Refs #1998 #2014 #1902 #1728

### DIFF
--- a/changelog.d/1998-nightly-residual-not-a-rate.removed.md
+++ b/changelog.d/1998-nightly-residual-not-a-rate.removed.md
@@ -38,12 +38,20 @@ entirely the **U** increment (at `Nx=40`, iteration 29: `err_U` 4.238e-04 agains
 while `test_fdm_upwind_stable`'s mass-conservation and `M.min() > 0` assertions constrain M's
 physics. The two sets are incomparable, not ordered, and that test stops at iteration 20.
 
-The rest is smaller than first stated. The `Nx=40` leg duplicates `test_fdm_upwind_stable` on every
-listed parameter, differing only in `max_iterations` (30 vs 20) and tolerance. And the coarse leg is
-not uncovered in kind: `tests/integration/test_fvm_hjb_coupling.py`'s `fdm_1d` fixture runs a coupled
-1D no-flux `FDM_UPWIND` solve at `Nx=25, Nt=12, sigma=0.4` asserting convergence, a 10× error drop,
-mass and positivity — strictly stronger, at comparable coarseness. What has no counterpart is the
-**low-σ / high-Péclet** coarse regime.
+The rest is smaller than first stated: the `Nx=40` leg duplicates `test_fdm_upwind_stable` on every
+listed parameter, differing only in `max_iterations` (30 vs 20) and tolerance.
+
+**The coarse leg has no counterpart anywhere, and an earlier version of this fragment said
+otherwise.** It claimed `tests/integration/test_fvm_hjb_coupling.py`'s `fdm_1d` fixture asserts
+"convergence, a 10× error drop, mass and positivity — strictly stronger". Those four assertions are
+made on the **`fvm_1d`** fixture, which runs `FVM_MUSCL`. `fdm_1d` runs `FDM_UPWIND` and has exactly
+one consumer, `test_1d_fvm_fdm_agreement`, which asserts cross-scheme agreement (`rel_m < 0.07`) and
+nothing about convergence, mass or positivity. Verified: `grep -n "fdm_1d\|fvm_1d"` over that file
+returns 9 lines, 3 of them the `fvm_1d`-only tests, 1 the shared agreement test.
+
+So a raise or a NaN from the coarse `FDM_UPWIND` solve is now caught by nothing — which is what this
+fragment originally said before the false citation replaced it. The **low-σ / high-Péclet** coarse
+regime is uncovered, and so is the coarse regime generally.
 
 **The case for repairing instead** is stronger than "fix the two-grid rate test", and worth weighing:
 `test_fdm_upwind_stable` already runs the identical `Nx=41, Nt=20, sigma=0.1, FDM_UPWIND` solve and
--- a/changelog.d/1998-nightly-residual-not-a-rate.removed.md
+++ b/changelog.d/1998-nightly-residual-not-a-rate.removed.md
@@ -1,0 +1,31 @@
+`TestDualityConvergence::test_mesh_refinement_improves_accuracy` is removed rather than repaired
+(#1998). It is the whole of "Nightly full test suite is failing": the validation shard reports
+`1 failed, 27 passed`, that one test, and every other shard is green.
+
+It recorded `result.max_error` — the final **Picard residual** — and compared two of them across
+`Nx = 20 / 40` under the name "h-convergence". Measured, with the test's own `tolerance=1e-8` and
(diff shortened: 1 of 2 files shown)


derrring:
Sixth review, independent and adversarial. **BLOCKERS-FOUND**, two, both fixed in `73c607f2` with
the gate GREEN.

**B1 — a false citation that overwrote a true statement.** The account claimed
`test_fvm_hjb_coupling.py`'s `fdm_1d` fixture asserts "convergence, a 10× error drop, mass and
positivity — strictly stronger, at comparable coarseness". Those four assertions are on the
**`fvm_1d`** fixture, which runs `FVM_MUSCL`. `fdm_1d` runs `FDM_UPWIND` and has exactly one
consumer — `test_1d_fvm_fdm_agreement` — asserting cross-scheme agreement (`rel_m < 0.07`) and
nothing else.

I verified it rather than taking the report: `grep -n "fdm_1d\|fvm_1d"` returns 9 lines against a
control of 5 test functions; `:128` and `:144` take `fvm_1d` alone, `:163` takes both, and the two
fixtures differ only in the scheme.

What makes this worse than a wrong pointer: `6e5e773b` had already written the **true** statement —
"the COARSE leg has no counterpart anywhere, so a raise or NaN from that specific solve is now caught
by nothing" — and `0db9af28` replaced it with this citation, shrinking the stated cost of the
deletion from *uncovered* to *not uncovered in kind*. The cost accounting was resting on it. Restored
in both the comment and the changelog fragment, with the measurement attached.

**B2 — the PR body was still v1.** It carried four claims that commits 2–6 of this branch identify as
false, and none of the corrections: "no commit in the green-to-red window changed the computation",
the wrong sibling precedent, "Nothing is lost", "strictly stronger". Nothing on the page marked it
superseded.

This is precisely what `6f63c22e` was written to fix — "the correction landed in two of three places;
the third said the opposite" — recurring in a fourth place I had not checked, and the one a merger
actually reads. The prescribed grep was run across the file, the fragment and the issue title; the PR
body is not a file, so it was outside the sweep's population. Body rewritten with a header marking
the replacement and a correction table covering all six rounds.

Review cost: 115k tokens, 43 tool calls, one agent.

Issue #1991, opened 2026-08-17
[Design] HJB solvers without a source_term channel could not be reached by MMS, so their convergence order was unmeasured


derrring:
`BaseHJBSolver.solve_hjb_system` takes `source_term: Callable | None = None`. Seven of the ten concrete HJB solvers override the method and **do not carry that parameter through**. Measured at `86ee6897` by `inspect.signature` over the class `__dict__`, not by grep:

| solver | accepts |
|---|---|
| `HJBFDMSolver` | `source_term`, `volatility_field` |
| `PenaltyHJBSolver` | `source_term`, `volatility_field` |
| `HJBFEMSolver` | inherits the base signature |
| **`HJBWENOSolver`** | `volatility_field` only |
| **`HJBGFDMSolver`** | `volatility_field` only |
| **`HJBSemiLagrangianSolver`** | `volatility_field` only |
| **`WeakFormHJBSolver`** | `volatility_field` only |
| **`NetworkHJBSolver`** | `volatility_field` only |
| **`NetworkPolicyIterationHJBSolver`** | `volatility_field` only |
| **`MeshlessGalerkinHJBSolver`** | neither |

## Why this is not an interface nit

A source term is how the **method of manufactured solutions** works: pick an exact `u`, feed the PDE residual `S = H(∇u)` back as a source, and the scheme must reproduce `u`. It is the only *external* oracle a numerical scheme has — everything else compares one implementation against another, and this repository has repeatedly measured implementations agreeing while both are wrong.

`tests/integration/test_mms_validation.py::TestMMSHJB1D::test_backward_heat_periodic_convergence` is the repo's HJB convergence test. It uses `HJBFDMSolver`. **That is not a choice — it is the only HJB solver reachable that way**, together with `PenaltyHJBSolver` and `HJBFEMSolver`.

So the observed convergence order of `HJBWENOSolver`, `HJBGFDMSolver`, `HJBSemiLagrangianSolver` and the rest **has never been measured, and cannot be measured through the public method**. Hit while trying to answer whether a high-order Neumann ghost (#1936) would lift the WENO wall's order: the measurement could not be set up.

```
TypeError: HJBWENOSolver.solve_hjb_system() got an unexpected keyword argument 'source_term'
```

## Related measurement

On 2026-08-17 the HJB wall was measured on a symmetric fixture, `sigma=0.3`, `Nt=20`:

| solver | \|∂ₙu\| at x_max, Nx=41 → 321 | |
[...]

PR #2016, opened 2026-08-19
test: coupled 2D MMS on a no-flux box, through the production iterator

derrring:
## The gap

`test_coupled_mfg_mms.py` measures coupled EOC in **1D** under **periodic** BC — deliberately, per
its own note, "to avoid the no-flux conservative-Laplacian boundary handling (#1075) so the measured
error" is the interior scheme. So coupled convergence had never been measured in 2D, nor with a wall
present at all.

## The pair — the paper's, with two deliberate changes

The GFDM paper's `eq:mms_reference` / `eq:mms_system` on $\Omega=(0,20)^2$, $T=4$, $\sigma=1$,
$\zeta=1/2$, with $c = \pi/L$ instead of $2\pi/L$ and an asymmetry $\beta = 0.6$ between the axes.
Both changes are there because the verbatim pair **cannot express the defects this test exists to
catch**, and both were verified by mutation:

| what the change buys | verbatim pair | this pair |
|---|---|---|
| $c=\pi/L$: the wall closure is real, not the periodic ghost | swapping the whole BC family to `periodic_bc` left every assertion green | `eu` $3.0125\text{e--}01 \to 1.2939\text{e+}01$ at $N_x=11$ (42×), EOC u collapses to $-0.118 / 0.067$ |
| $\beta$: the fixture can express a wrong-axis read | $u$, $m$ and both sources exactly transpose-symmetric, so `eu` from `U[0]` and `U[0].T` is bit-identical | `eu` from `U[0].T` is 45.7× / 81.8× / 119.5× the correct one |

$c=\pi/L$ still gives $\sin(cL)=0$, so the pair stays exactly Neumann-compatible.

Sources re-derived for **this** pair symbolically (sympy, from the definitions of $u$ and $m$) and
cross-checked against the hand-written vectorised forms in the file to **5.6e-17 / 1.1e-19** over 400
random $(t,x)$, against residual scales 2.3e-01 / 4.5e-04. The comparison can fail: perturbing
$r_m$ by $10^{-7}$ relative registers as 4.5e-11.

Two properties the paper states also hold here, verified symbolically on this pair:

- $\int_\Omega r_m\,dx = 0$ exactly
- $\partial u/\partial x_k = \partial m/\partial x_k = 0$ on each of the four walls

The second is load-bearing: $\alpha = -\nabla u$, so $\alpha\cdot n = 0$ and hence
$J\cdot n = \alpha m - D\nabla m = 0$. **The pair is exactly compatible with the no-flux wall**,
which is what lets it run on this box.

## Measured

| $N_x$ | $N_t$ | Picard | `eu` | EOC u | `em` | EOC m |
|---|---|---|---|---|---|---|
| 11 | 8 | 4 | 3.0125e-01 | — | 9.876e-03 | — |
| 21 | 16 | 4 | 1.5759e-01 | 0.935 | 5.036e-03 | 0.972 |
| 31 | 24 | 4 | 1.0534e-01 | **0.993** | 3.383e-03 | **0.982** |

47.5 s, `@pytest.mark.slow` + `@pytest.mark.integration`, matching the existing coupled MMS test's
markers.

## Two disciplines

**`assert result.converged` before any error metric.** This is what #1998 turned out to be: a test
comparing two unconverged Picard residuals under the name "h-convergence". Here it is cheap
insurance rather than a live hazard — measured, this family converges at every $\sigma$ tried, in 4
iterations each time:

| $\sigma$ | 1.0 | 0.5 | 0.3 | 0.1 |
|---|---|---|---|---|
| `eu` at $N_x=31$ | 1.0534e-01 | 6.8685e-02 | 6.4407e-02 | 6.3901e-02 |
| converged | yes | yes | yes | yes |

**An error LEVEL beside the order.** An order alone is a min over the HJB scheme, the FP scheme and
the wall closure. Bounds from measurement — $\sigma$ wrong at problem **construction** while the
sources stay closed over the true value, so only the discretized PDE is wrong:

| $D$ factor | `eu`(11) | `eu`(21) | `eu`(31) | EOC u |
|---|---|---|---|---|
| 1.00 | 3.0125e-01 | 1.5759e-01 | 1.0534e-01 | 0.935, 0.993 |
| 1.21 | 3.5252e-01 | 2.4572e-01 | 2.2015e-01 | 0.521, 0.271 |
| 2.00 | 9.6882e-01 | 9.5405e-01 | 9.5607e-01 | 0.022, −0.005 |

`eu[-1] < 1.5e-01` sits 1.42× above baseline and 1.47× below the 1.21× mutant. **`em` is not the
discriminating field** for this defect — at 1.21× it moves only 3.383e-03 → 3.487e-03 and its EOC
stays inside the band at 0.956 / 0.945 — so no level bound is asserted on it.

## What it cannot see — in the file, not just here

- **The wall is tangential.** $\partial_\nu u = 0$ makes $\alpha\cdot n = 0$, so this exercises
  compatibility and not the treatment of a normal drift. That is #2006, FP-only.
- **It does not measure the coupling direction**, and the paper says so of its own instance.
  Measured on this pair: $\zeta m$ has RMS 1.26e-03 — 8.1% of the $|\nabla u|^2/2$ term (1.55e-02)
  and 1.1% of the whole HJB residual (1.17e-01, dominated by $-\partial_t u$ at 1.03e-01). Against
  that, the discretization error at the finest level is 0.41% relative in $u$ and 6.6% in $m$.
## Which solvers can run this at all — the prior question to convergence order

Measured at $N_x=21$ on this exact fixture, each candidate paired against FDM on the other side so a
failure localises. **2 of 8 pairings run.**

| HJB × FP | outcome |
|---|---|
| FDM × FDM | converged, 4 Picard iterations, `eu` = 1.5759e-01 |
| GFDM × FDM | converged, 4 Picard iterations, `eu` = **1.9699e-02** |
| WENO × FDM | `NotImplementedError`: `source_term` is threaded through the 1D path only; this is 2D |
| SemiLagrangian × FDM | `NotImplementedError`: `solve_hjb_system` does not accept `source_term` |
| FDM × GFDM / Particle / SL | `NotImplementedError`: `solve_fp_system` does not accept `source_term` |
| FDM × FVM | `ValueError`: cannot broadcast (21,21) against (882,) |

So the question splits, and only the second half is about convergence:

- **Reachability.** Six of eight never enter the solve. Not non-convergence — an MMS needs a source
  channel and most solvers have none. Filed: #2020 (FP side), #1991 (HJB side), #2019 (the FVM
  crash, which is a source-argument *convention* mismatch: FVM passes `geometry.meshgrid()`,
  a `(2,21,21)` tuple, where every other caller passes the `(441,2)` point list).
- **Convergence.** Both reachable pairings converge in 4 iterations, neither marginally. GFDM is 8×
  more accurate than FDM at the same grid.

The earlier wording — "the pair is method-agnostic; other schemes are separate parametrisations" —
read that structural gap as a matter of adding parametrisations. It is not.

GFDM's source was verified to reach it rather than assumed from its signature: zeroing $s_{HJB}$
moves its `eu` from 1.9699e-02 to 9.7640e+00, a factor of 496.

## Corrections to earlier versions of this PR

Each was stated here with numbers, and each was wrong:

1. **"Two properties … independent confirmation of the transcription."** Neither is. $\int r_m = 0$
   is one scalar per time and is blind to anything that integrates away — an $r_m$ carrying a
   spurious $\tfrac17\cos(2cx_1)\cos(4cx_2)$ integrates to exactly 0 as well (checked symbolically).
   The wall-derivative property is a fact about $u$ and $m$ alone and never touches the sources.
   Only the sympy cross-check bears on the transcription.
2. **"$\alpha = -$`coupling_coefficient`$\cdot\nabla U$."** Wrong source, right number. The scale
   comes from `fp_drift_coefficient` (`utils/pde_coefficients.py:118`), which for a
   quadratic-MINIMIZE `SeparableHamiltonian` returns $1/\lambda$ and never reaches the
   `coupling_coefficient` fallback (#1420 / G-017). $\lambda = 1$ here, so both are 1.0 — which is
   exactly why the misattribution left no trace.
3. **"This problem family does not converge at $\sigma=0.1$."** That measurement belongs to the 1D
   duality family of #1998 and was transferred here by label. Measured on this family: it converges
   at $\sigma = 0.1$ (table above).
4. **"$\zeta m$ is ~2.5% of the HJB residual (1.28e-03 against 1.24e-01)"** and **"~30% relative
   discretization error."** Both were measured on the verbatim pair, and the 1.24e-01 was labelled
   as the $|\nabla u|^2/2$ term when it was the total. Corrected figures above.
5. **The discrimination bound `9e-01`** came from the verbatim pair. On this one it would have
   **passed the 1.21× mutant outright** (2.20e-01) and cleared 2× by only 6%.
6. **The earlier `_eoc` divided by point counts, not the mesh ratio** — every order inflated 7.2% /
   4.1%, while `_solve` twelve lines up computed `dx` with `nx - 1` correctly. Fixed in `172b2bae`.

## A trap worth naming: `problem.sigma` is inert after construction

Re-deriving the bound, the first attempt mutated `problem.sigma` **after** building the problem and
asserted `problem.diffusion == k·D`, which passed. The solves came back **byte-identical** at
$k=1.21$ and $k=2.0$ — every printed digit. `get_diffusion_coefficient_field`
(`mfg_problem.py:1416`) resolves override → `volatility_field` → `self.sigma`, and
`volatility_field` is snapshotted at construction (`:560`), so the scalar is the last-resort
fallback and nothing reads it. Meanwhile `problem.diffusion`, whose docstring says it "is the
coefficient appearing directly in the PDE", keeps reporting the mutated value. The assert verified
one level below the claim. Filed separately.

## Gate

`./scripts/local_ci.sh`: **GATE GREEN**, full suite 6292 passed / 12 skipped / 23 xfailed in 140 s.
The test itself: 1 passed in 47.5 s.

Refs #1728 #2006 #1998 #1420

### DIFF
--- a/changelog.d/coupled-2d-mms-no-flux.added.md
+++ b/changelog.d/coupled-2d-mms-no-flux.added.md
@@ -4,25 +4,50 @@ A coupled 2D MMS on a no-flux box, run through the production `FixedPointIterato
 its own note, to avoid the no-flux boundary handling so the measured error is the interior scheme.
 So coupled convergence had never been measured in 2D, nor with a wall present at all.
 
-The manufactured pair is the GFDM paper's `eq:mms_reference` / `eq:mms_system` verbatim, on
-`Ω = (0,20)²`, `T = 4`, `σ = 1`, `ζ = 1/2`. Sources derived symbolically and cross-checked against a
-hand-written vectorised form to `1.1e-16` / `3.3e-19` over 200 random `(t,x)`. Two properties the
-paper states are reproduced as independent confirmation of the transcription: `∫_Ω r_m dx = 0`
-exactly, and `∂u/∂x_k = ∂m/∂x_k = 0` on every wall — the second is what makes `α·n = 0`, hence
-`J·n = 0`, so the pair is exactly compatible with the no-flux wall.
+The manufactured pair is the GFDM paper's `eq:mms_reference` / `eq:mms_system` with **two deliberate
+changes**, on `Ω = (0,20)²`, `T = 4`, `σ = 1`, `ζ = 1/2`. It is not the paper's pair verbatim, and
+both changes exist because the verbatim one cannot express the defects this test is for:
 
-Measured over `Nx = 11/21/31` with `Nt ∝ Nx`: EOC `u` 0.95 → 1.01, `m` 0.99 → 1.00, Picard
-converging in 4 outer iterations at every level. 48.9 s, marked `slow` and `integration`.
+- `c = π/L` instead of `2π/L`. Still Neumann-compatible (`sin(cL) = 0` either way) but no longer
+  L-periodic, so the mirror ghost and the periodic ghost stop coinciding. With the paper's `c`,
+  swapping the entire boundary-condition family to `periodic_bc` left every assertion green; with
+  this one it moves `eu` from `3.0125e-01` to `1.2939e+01` and collapses EOC `u` to `−0.118 / 0.067`.
+- `β = 0.6` breaking the symmetry between the axes, plus a `2c/4c` mix in `m`. The paper's pair is
+  exactly transpose-symmetric, so `eu` from `U[0]` and from `U[0].T` is bit-identical and a
+  wrong-axis read is unexpressible; here the transposed read is 45.7× / 81.8× / 119.5× off.
 
-The test asserts an error **level** beside the order, because an order alone is a min over the HJB
-scheme, the FP scheme and the wall closure. Bounds from measurement: baseline `eu` at the finest
-level is `5.959e-01`; a solver whose `D` is off by 1.21× gives `1.326e+00` with EOC `u` collapsing
-1.01 → 0.38, and by 2× gives `4.465e+00` with EOC → 0.06. Verified mutation-red against the 1.21×
-case.
+Sources re-derived for **this** pair with sympy and cross-checked against the hand-written
+vectorised forms to `5.6e-17` / `1.1e-19` over 400 random `(t,x)`, against residual scales `2.3e-01`
+/ `4.5e-04`, with a positive control (a `1e-7` relative perturbation registers at `4.5e-11`).
 
-Two limits are stated in the file rather than left to be discovered. The wall is **tangential**
-(`∂_ν u = 0`), so this exercises compatibility and not the treatment of a normal drift — that is
-#2006, FP-only. And it does not measure the **coupling direction**: `ζm` is ~2.5% of the HJB residual
-(RMS `1.28e-03` against `1.24e-01`), which the paper says of its own instance too. A model-side
-perturbation of `ζ` is swamped by the ~30% relative discretization error at these resolutions; a
-solver-side error is not, which is what the discrimination above measures.
+Two properties the paper states also hold for this pair — `∫_Ω r_m dx = 0` exactly, and
+`∂u/∂x_k = ∂m/∂x_k = 0` on all four walls. **Neither is a check on the transcription**: the first is
+one scalar per time and is blind to anything that integrates away, the second never touches the
+sources. The second matters for a different reason — it makes `α·n = 0`, hence `J·n = 0`, so the pair
+is exactly compatible with the no-flux wall, which is what lets it run on this box.
+
+Measured over `Nx = 11/21/31` with `Nt ∝ Nx`: `eu` `3.0125e-01 → 1.5759e-01 → 1.0534e-01`, EOC `u`
+0.935 → 0.993; `em` `9.876e-03 → 5.036e-03 → 3.383e-03`, EOC `m` 0.972 → 0.982. Picard converges in
+4 outer iterations at every level, and at every `σ` tried (1.0, 0.5, 0.3, 0.1). 47.5 s, marked
+`slow` and `integration`.
+
+Three limits are stated in the file rather than left to be discovered.
+
+**It cannot resolve a coefficient error below ~10%.** Measured across two solver-side defect
+families: a 5% error in the diffusion coefficient or in the drift scale passes every assertion. The
+error-level assertion is kept as a regression guard on the constant — EOC is a ratio and is blind to
+a uniform scaling — but it is *not* a discriminator: over ten measured mutants it never fails
+without the EOC assertion failing first.
+
+**The wall is tangential** (`∂_ν u = 0`), so this exercises compatibility and not the treatment of a
+normal drift — that is #2006, FP-only.
+
+**It does not measure the coupling direction**, which the paper says of its own instance too. `ζm`
+has RMS `1.26e-03` — 8.1% of the `|∇u|²/2` term and 1.1% of the whole HJB residual, which is
+dominated by `−∂_t u`. The discretization error at the finest level is 0.41% relative in `u` and
+6.6% in `m`, so a model-side perturbation of `ζ` is not resolvable here.
+
+Separately, the study established which solvers can run an MMS at all on this box: **2 of 8 solver
+pairings**. `FDM × FDM` and `GFDM × FDM` converge; the rest never enter the solve because
+`solve_*_system` does not take a `source_term` (#1991, #2020) or crash on the source argument
+convention (#2019).
--- a/changelog.d/coupled-2d-mms-no-flux.added.md
+++ b/changelog.d/coupled-2d-mms-no-flux.added.md
@@ -1,0 +1,28 @@
+A coupled 2D MMS on a no-flux box, run through the production `FixedPointIterator`.
+
+`test_coupled_mfg_mms.py` measures coupled EOC in **1D** under **periodic** BC — deliberately, per
+its own note, to avoid the no-flux boundary handling so the measured error is the interior scheme.
+So coupled convergence had never been measured in 2D, nor with a wall present at all.
+
+The manufactured pair is the GFDM paper's `eq:mms_reference` / `eq:mms_system` verbatim, on
+`Ω = (0,20)²`, `T = 4`, `σ = 1`, `ζ = 1/2`. Sources derived symbolically and cross-checked against a
+hand-written vectorised form to `1.1e-16` / `3.3e-19` over 200 random `(t,x)`. Two properties the
+paper states are reproduced as independent confirmation of the transcription: `∫_Ω r_m dx = 0`
+exactly, and `∂u/∂x_k = ∂m/∂x_k = 0` on every wall — the second is what makes `α·n = 0`, hence
+`J·n = 0`, so the pair is exactly compatible with the no-flux wall.
+
+Measured over `Nx = 11/21/31` with `Nt ∝ Nx`: EOC `u` 0.95 → 1.01, `m` 0.99 → 1.00, Picard
+converging in 4 outer iterations at every level. 48.9 s, marked `slow` and `integration`.
+
+The test asserts an error **level** beside the order, because an order alone is a min over the HJB
+scheme, the FP scheme and the wall closure. Bounds from measurement: baseline `eu` at the finest
+level is `5.959e-01`; a solver whose `D` is off by 1.21× gives `1.326e+00` with EOC `u` collapsing
+1.01 → 0.38, and by 2× gives `4.465e+00` with EOC → 0.06. Verified mutation-red against the 1.21×
+case.
+
+Two limits are stated in the file rather than left to be discovered. The wall is **tangential**
+(`∂_ν u = 0`), so this exercises compatibility and not the treatment of a normal drift — that is
+#2006, FP-only. And it does not measure the **coupling direction**: `ζm` is ~2.5% of the HJB residual
+(RMS `1.28e-03` against `1.24e-01`), which the paper says of its own instance too. A model-side
+perturbation of `ζ` is swamped by the ~30% relative discretization error at these resolutions; a
+solver-side error is not, which is what the discrimination above measures.
--- a/tests/integration/test_coupled_mms_2d_no_flux.py
+++ b/tests/integration/test_coupled_mms_2d_no_flux.py
@@ -157,10 +157,15 @@ def _solve(nx: int, nt: int):
 
 
 def _eoc(errors, levels):
-    return [
-        float(np.log(errors[i] / errors[i + 1]) / np.log(levels[i + 1][0] / levels[i][0]))
-        for i in range(len(errors) - 1)
-    ]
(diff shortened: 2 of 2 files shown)


derrring:
Converting to draft. Four blockers, one of them a code bug that changes every number, and a mutation
report showing the wall is not exercised at all.

## B1 — the EOC denominator was the point count, not the mesh ratio

`_eoc` divided by `levels[i+1][0] / levels[i][0]`. But `dx = L/(nx-1)` — which `_solve` twelve lines
up computes correctly — so refining 11 → 21 → 31 gives mesh ratios **2.0 and 1.5**, not 21/11 = 1.909
and 31/21 = 1.476. **The file disagreed with itself.** Every order was inflated by 7.2% and 4.1%:

| | reported | true |
|---|---|---|
| EOC u | 0.95 / 1.01 | **0.885 / 0.966** |
| EOC m | 0.99 / 1.00 | **0.922 / 0.958** |

Fixed and re-measured; the true figures reproduce the reviewer's to three digits. The assertion band
`[0.8, 1.3]` was really `[0.746, 1.213]` in true order, so a scheme degraded to true order 0.75 would
have reported 0.804 and **passed** a test whose message says "u is not first order".

## B2 — my "two independent confirmations of the transcription" confirm nothing about it

`∫_Ω r_m dx = 0` survives **all ten** plausible `s_fp` transcription errors the reviewer tried — sign
flip on diffusion, sign flip on the whole advection, factor 2 on advection, dropping `m·Δu`,
`d_t m` sign flipped, `a2'` wrong by 3, `lap_m` halved, `x1↔x2` swapped, `σ²` dropped entirely. It is
forced by the ansatz (∫m is time-constant; every other term is a divergence killed by the wall
property), not by the derivation. The wall-derivative check touches only `u` and `m`.

**And neither check looks at `s_hjb` at all.** So a sign error on the HJB source ships with both
"confirmations" green, and the test then measures clean first-order convergence to a *different*
exact solution — precisely the failure the docstring says they rule out. The sympy cross-check is the
real evidence; those two were rhetoric.

## B3 — `coupling_coefficient=1.0` is bit-identically inert, and AGENTS.md already says so

Measured: `1.0`, `0.5`, `0.0`, `-3.0` all give `eu=1.6283294993e+00`, `em=8.5802571283e-03`.
`fp_drift_coefficient` returns `1/control_cost.lambda_` for a quadratic-MINIMIZE
`SeparableHamiltonian` and never reaches the `coupling_coefficient` lookup. What actually pins
`α = -∇u` is `QuadraticControlCost(control_cost=1.0)`.

`AGENTS.md:97` states this outright. I inherited the stale comment from `test_coupled_mfg_mms.py`
without checking it — and `changelog.d/1442-coupled-fixture-drift.fixed.md` records six tests
breaking from exactly this misattribution (#1420 / G-017).

## B4 — "this problem family does not converge at σ=0.1" does not reproduce

Measured on **this** fixture at every level, both readings (σ changed in the solver only, and σ
changed consistently so the sources move with it): `converged=True` in 4–5 iterations throughout.

The non-convergence I measured was on `test_duality_convergence`'s `_default_components` fixture — a
**different problem**. I generalised across problems. The `assert result.converged` discipline is
still right and the guard is correctly placed (verified: forcing `max_iterations=1` raises its own
message), but the sentence justifying it names a population it was not measured on.

## The mutation report is the part that needs design, not editing

| mutant | result |
|---|---|
| **`periodic_bc` replacing the no-flux wall, identical pair** | **all assertions GREEN** |
| **axis transpose** | **structurally invisible** |
| `divergence_centered` (2nd-order FP advection) | GREEN — `eu` moves 0.03% |
| D × 1.21 | red, as claimed |

The pair is even about every wall **and** L-periodic, so the mirror ghost and the periodic ghost
coincide: swapping the entire BC family leaves the test green. "Nothing measured coupled convergence
with a wall present" describes the configuration; as *coverage* the wall closure is undiscriminated.

And the transpose blindness is worse than a gap — measured, `U`, `M`, `u_star`, `m_star`, `s_hjb`
and `s_fp` are **all exactly transpose-symmetric**, so `eu` from `U[0]` and from `U[0].T` are
bit-identical. This fixture cannot express "read the potential along the wrong axis" — which is one
of the three mutants **#2017, same author, same day**, cites as the reason to go to 2D.
`test_fp_mms_wall_order_1728.py:112` already knows the fix (`A_FULL = [0.7, -0.4, 0.2]`,
"distinctness breaks a transposed reshape"); this fixture does not use it.

Also: `eu` moves 0.03% under an FP-scheme swap while `em` moves 4.7% — and only `eu` has a level
assertion. The level guards the wrong field.

## Smaller, all mine

- "~2.5% of the HJB residual (RMS 1.28e-03 against 1.24e-01)" — those two numbers give 1.03%, not
  2.5%, and they are sampled on different domains (1.24e-01 is the t=0-only RMS). Consistently
  sampled it is **1.41%**.
- "~30% relative discretization error" — measured 1.89–4.69% (u), 5.94–15.60% (m). Nothing reaches 30%.
- The `9e-01` level bound is not pinned to `LEVELS`; the 1D sibling asserts exactly that guard and I
  omitted it here.
- `o <= 1.3` is **inert** — no available scheme change lifts the overall order, so nothing can trip it.
- `_split`'s `else` branch is a fail-silent fallback that does not fire today.
- "GATE GREEN" cited `--fast`, which prints `SKIPPED test suite`. It is not evidence about this file.

Rework: fix the four blockers, break the transpose symmetry, and decide whether the wall can be
exercised at all by a pair that is periodic-compatible — or state plainly that it cannot and rename
what the test claims.

derrring:
Body rewritten rather than patched: six claims in it were wrong, not merely stale, and leaving them
struck inline would have made the record harder to read than the corrections are worth. What changed,
so the edit is not silent:

1. **"Two properties … independent confirmation of the transcription."** Neither is. An $r_m$ with a
   spurious $\tfrac17\cos(2cx_1)\cos(4cx_2)$ integrates to exactly 0 as well — checked symbolically.
   Only the sympy cross-check bears on the transcription, and it was re-run for the current pair:
   5.6e-17 / 1.1e-19 over 400 points, with a positive control.
2. **The drift scale is not `coupling_coefficient`.** `fp_drift_coefficient` returns $1/\lambda$ and
   never reaches that fallback. Both are 1.0 here, which is why it left no trace.
3. **"This family does not converge at $\sigma=0.1$."** Transferred by label from #1998's 1D duality
   family. Measured here: converges at $\sigma \in \{0.1, 0.3, 0.5, 1.0\}$, 4 iterations each.
4. **The $\zeta m$ fraction and the "~30% relative discretization error"** were measured on the
   verbatim pair, and the 1.24e-01 was labelled as the $|\nabla u|^2/2$ term when it was the total.
5. **The bound `9e-01`** would have passed the 1.21× diffusion mutant outright on this fixture.
   Re-derived to 1.5e-01 from a measured table.
6. **"Other schemes are separate parametrisations."** Six of eight solver pairings cannot run this
   at all. That is a structural gap, now filed as #2019 / #2020.

Two of those were caught only because the check was aimed at the claim rather than one level below
it. The bound re-derivation first mutated `problem.sigma` after construction and asserted
`problem.diffusion == k·D` — the assert passed and the solves came back **byte-identical** at 1.21×
and 2.0×. `get_diffusion_coefficient_field` resolves override → `volatility_field` → `self.sigma`,
and `volatility_field` is snapshotted at construction, so the scalar is the last-resort fallback and
nothing reads it. Supplying the wrong $\sigma$ at construction is what makes the mutation land.

Gate on this revision: **GATE GREEN**, 6292 passed / 12 skipped / 23 xfailed in 140 s.

derrring:
Independent adversarial review found four blockers. All four are now fixed in `cf70dae7`; the gate
is GREEN (6292 passed / 12 skipped) and the test runs in 44.6 s.

Worth stating what the review did **not** find, because it changes how to read the rest: the
docstring's measured numbers are honest. The reviewer reproduced the full 3×3 diffusion table, every
EOC, the transpose ratios, the term-budget RMS values and the relative errors, all to the printed
digits, and said explicitly that auditing those alone would have returned MERGE-OK. Every defect was
in a **claim about** the numbers, not in a number.

1. **The changelog fragment was never updated when the fixture changed.** It still described the
   paper's verbatim pair, quoted errors 3–6× wrong, and repeated *verbatim* the "independent
   confirmation of the transcription" sentence that `8eabbd93` exists to retract. The correction
   landed in the test docstring and not in the user-facing artifact one file over, so it did not
   actually land. Rewritten.

2. **"The previous bound of 9e-01 would have PASSED the 1.21× mutant" is false** — the EOC assertion
   rejects that mutant on 0.521 / 0.271 whatever the level bound is. Measured over **ten mutants in
   two independent families**:

   | | 1.05 | 1.10 | 1.15 | 1.21 | 2.00 |
   |---|---|---|---|---|---|
   | wrong diffusion, EOC-assert | pass | **FAIL** | FAIL | FAIL | FAIL |
   | wrong diffusion, level-assert | pass | pass | FAIL | FAIL | FAIL |

   | | 1.05 | 1.20 | 1.50 |
   |---|---|---|---|
   | wrong drift scale, EOC-assert | pass | FAIL | FAIL |
   | wrong drift scale, level-assert | pass | FAIL | FAIL |

   The level assertion never fails alone, and at k = 1.10 the EOC fails while it passes. Kept, but
   relabelled as a **regression guard on the constant** — EOC is a ratio and cannot see a uniform
   scaling — and explicitly marked unexercised rather than sold as discrimination.

   The same sweep found the **detection floor**, now recorded in WHAT THIS STUDY CANNOT SEE: a 5%
   error in *either* coefficient passes the entire test. Any order this test reports is an order at
   that sensitivity.

3. **`_split`'s else-branch was a fail-silent fallback** aliasing `x2 := x1`, which would evaluate
   both manufactured sources on the degenerate diagonal instead of raising — in an oracle, the one
   failure mode that cannot be afforded. Instrumented over a full converged solve it was taken 0 of
   64 times. Now raises, and names #2019, since the FVM argument convention is exactly the shape
   that would have reached it.

4. **The inline comment on `coupling_coefficient=1.0` re-asserted the misattribution** the docstring
   corrects two screens above. Verified inert: solves at 0.5 / 1.0 / 7.0 are bit-identical,
   `max|dU| = max|dM| = 0.000e+00`, against a control where `sigma=1.1` moves the same solve by
   `1.672e-02`.

The reviewer's own note on scope is worth keeping: it was asked to review the test file, and
`git diff --stat main...HEAD` shows the branch touches two — the second is where the worst defect
was. On this repo a `changelog.d` fragment is a claim-bearing artifact written at the start of a
branch and rarely re-read after the branch changes shape underneath it.

Review cost: 103k tokens, 34 tool calls, one agent, read-only against the working tree.

PR #2017, opened 2026-08-19
docs: the dimension must be able to express the property under test

derrring:
Adds one rule to `AGENTS.md` § Testing.

## The rule

Not "prefer 2D" and not "1D is cheaper":

> **A test in a dimension that cannot express the property is not a weak test — it cannot fail at
> all**, and it reads exactly like a passing one.

- **Needs d ≥ 2**: a normal/tangential decomposition, an axis pairing, a corner, a transpose,
  anisotropy. A 1D wall's normal is always the coordinate axis and there is no tangential component.
- **Fully expressed in d = 1**: sign conventions, time-stepping order, convergence criteria, scalar
  contracts, source-term plumbing.

## The discriminator is measurable, so it need not be argued

Write a mutation that breaks the property and run it in 1D:

```
drop the tangential advection at wall rows   1D: max|diff| = 0.000e+00   2D: separates
flip its sign                                1D: max|diff| = 0.000e+00   2D: separates
read the potential along the wrong axis      1D: max|diff| = 0.000e+00   2D: separates
```

Three mutants, three exact zeros, measured on #2006's FP wall study. `max|diff| = 0` in 1D means the
test must go up a dimension.

## The burden runs both ways

2D costs real time — the same FP MMS study is **0.1 s at d=1 against 1.5 s at d=2**. So a 2D test
whose 1D reduction separates the same mutants owes its runtime an explanation, and a 1D test of a
directional property owes a mutant showing it can fail.

This is the cross-project "non-discriminating test data" rule applied to dimension — the same shape
as a uniform density that cannot separate a gradient-form bug from a correct scheme.

## Why a rule rather than a preference

A preference ("2D unless you must") puts the burden on justifying 1D, which is an argument. This puts
the burden on a measurement that either party can run in a minute, and it cuts both ways.

`CLAUDE.md` is a symlink to `AGENTS.md`, so it is covered.

Gate: `./scripts/local_ci.sh --fast` **GREEN**.

Refs #1728 #2006

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -131,6 +131,38 @@ Cross-project testing discipline governs **what** a test must cover (edge/stress
 | Visualization | Sometimes | Yes | Smoke |
 | Utility function | No | Internal | Unit or smoke |
 
+### The dimension must be able to express the property under test ⚠️
+
+Not "prefer 2D" and not "1D is cheaper". The rule is that **a test in a dimension that cannot
+express the property is not a weak test — it cannot fail at all**, and it reads exactly like a
+passing one.
+
+- **Needs d ≥ 2**: anything involving a normal/tangential decomposition, an axis pairing, a corner,
+  a transpose, or anisotropy. A 1D wall's normal is always the coordinate axis and there is no
+  tangential component, so a scheme that mishandles the tangential part passes 1D by construction.
+- **Fully expressed in d = 1**: sign conventions, time-stepping order, convergence criteria, scalar
+  contracts, source-term plumbing. Here 2D buys runtime and nothing else — measured, the same FP
+  MMS study runs 0.1 s at d = 1 and 1.5 s at d = 2.
+
+**The discriminator is measurable, so do not argue it.** Write a mutation that breaks the property,
+run it in 1D, and read the difference:
+
+```
+drop the tangential advection at wall rows   1D: max|diff| = 0.000e+00   2D: separates
+flip its sign                                1D: max|diff| = 0.000e+00   2D: separates
+read the potential along the wrong axis      1D: max|diff| = 0.000e+00   2D: separates
+```
+
+Three mutants, three exact zeros (#1728/#2006). **`max|diff| = 0` in 1D means the test must go up a
+dimension**; a non-zero difference means 1D expresses it and the extra dimension is cost without
+coverage.
+
+The burden runs both ways: a 2D test whose 1D reduction separates the same mutants owes its runtime
+an explanation, and a 1D test of a directional property owes a mutant showing it can fail.
+
+This is the cross-project "non-discriminating test data" rule applied to dimension — the same shape
+as a uniform density that cannot separate a gradient-form bug from a correct scheme.
+
 ### Closing out a fix ⚠️ — name the oracle, or say there isn't one
 
 "Add a test" is **not** the default close-out for a fix here. Measured on this repo: the six load-bearing
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -140,25 +140,43 @@ passing one.
 - **Needs d ≥ 2**: anything involving a normal/tangential decomposition, an axis pairing, a corner,
   a transpose, or anisotropy. A 1D wall's normal is always the coordinate axis and there is no
   tangential component, so a scheme that mishandles the tangential part passes 1D by construction.
+  Note the scope: d ≥ 2 is **necessary** for these, not sufficient — see the symmetry measurement
+  below.
 - **Fully expressed in d = 1**: sign conventions, time-stepping order, convergence criteria, scalar
   contracts, source-term plumbing. Here 2D buys runtime and nothing else — measured, the same FP
   MMS study runs 0.1 s at d = 1 and 1.5 s at d = 2.
 
 **The discriminator is measurable, so do not argue it.** Write a mutation that breaks the property,
-run it in 1D, and read the difference:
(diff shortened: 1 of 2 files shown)


derrring:
Converting to draft. Three blockers, and the third is that **#2016 — my own PR, same day — is the
live counterexample to this rule as written.**

## B1 — the rule's table says 3/3 separate in 2D; every other artifact says 2/3

`AGENTS.md:151-153` renders all three mutants as `2D: separates`. The in-tree source it cites,
`test_fp_mms_wall_order_1728.py:32-35` — on main, predating this PR — says:

> "…are **exact no-ops in 1D**, max error difference 0.000e+00, and **two of the three separate in 2D**."

This PR's own changelog fragment also says "two of three". The merged rule is the only one of the
three artifacts claiming 3/3.

Not cosmetic, because the imperative built on it is *"`max|diff| = 0` in 1D means the test must go up
a dimension"*. On the mutant that does **not** separate in 2D, going up a dimension does not make the
test discriminate. `max|diff| = 0` in 1D establishes that **1D does not express it** — it says nothing
about whether **2D does**. The rule conflates those, and its own evidence refutes the conflation on
one of three cases.

## B2 — "needs d ≥ 2: anything involving a normal/tangential decomposition" over-quantifies

A decomposition needs d ≥ 2 to *have* a tangential part. A defect in the **normal** half is fully
expressible in 1D — and the file I cite proves it:
`test_divergence_upwind_is_first_order_at_a_drifting_wall[1-±1]` and
`test_gradient_upwind_does_not_converge_at_a_drifting_wall[1]` are 1D wall tests that discriminate
wall-normal defects, at 0.06–0.07 s each.

Read without the tangential sentence supporting it, the bullet sends every wall test to 2D, including
the ones the same file shows working in 1D.

## B3 — the rule omits what makes 2D actually express the property, and I shipped the counterexample

"Go up a dimension" is necessary, not sufficient. **#2016 is a 2D test whose pair, sources, domain
and grid are all exactly transpose-symmetric** — measured, `eu` from `U[0]` and from `U[0].T` are
bit-identical — so it cannot express *"read the potential along the wrong axis"*, the third row of
this very table.

`test_fp_mms_wall_order_1728.py:112` already practises the missing half: `A_FULL = [0.7, -0.4, 0.2]`,
*"distinctness breaks a transposed reshape"*. A rule whose stated hazard is *"it reads exactly like a
passing one"* has to carry that clause, or it licenses the same failure one dimension up — which is
what I did today, in the PR I opened alongside this one.

## Verified, and worth keeping

- **0.1 s vs 1.5 s** confirmed: `[1-1]` = 0.07 s, `[2-1]` = 1.22 s. If anything it understates the
  penalty, since the d=1 study runs three levels against d=2's two and still costs ~17× less.
- **The three zeros are traceable**, to `test_fp_mms_wall_order_1728.py:32-35` on main, predating this
  PR. The evidence is not gone — the problem is that the same sentence supplies "two of three" and
  that half did not make it into the rule.

## Non-blocking, and one of them matters for whether this rule should exist as written

`AGENTS.md:191-192` already requires *"A mutation-verified convention pin — the mutation and its kill
count stated"*, and `:201` *"Adding tests must be justified by a failure they discriminate"*. Both are
25–60 lines below, in the same section. The new rule's two burdens are dimension-specific instances of
both, stated without reference to either — in a repo that runs `check_single_source.py` against
exactly this shape.

Also: *"the discriminator is measurable, so do not argue it"* over-claims. Measurable is whether a
**given** mutant separates. Whether the property is "directional", and which mutant expresses it,
is upstream of the measurement and stays an argument. The rule automates the cheap half.

Rework: correct 3/3 → 2/3 and stop inferring "2D expresses it" from "1D does not"; scope the d ≥ 2
bullet to the tangential half; add the clause about breaking symmetry in the added dimension; and
decide whether this belongs as a standalone rule or as two sentences under the mutation-pin rule that
already exists.

derrring:
Out of draft. Two corrections, both from measurement rather than argument.

**The table overstated 2D.** It claimed all three tangential mutants "separate" in 2D. The merged
fixture they come from says otherwise in its own docstring — three exact zeros in 1D, and **two of
the three** separate in 2D. The section was quoting the half that supported it.

**That third mutant is the rule's missing half**, and today's work supplied a much sharper instance.
A 2D fixture that is symmetric in the relevant variable cannot express a defect in that variable
either, and it looks exactly as healthy as one that can. Measured on #2016, where the GFDM paper's
verbatim manufactured pair is transpose-symmetric and periodic on the box:

| mutant | verbatim 2D pair | after breaking the symmetry |
|---|---|---|
| swap the whole BC family, `no_flux` → `periodic` | **byte-identical** | `eu` 3.0125e-01 → 1.2939e+01 (42×), EOC → −0.118 |
| read the potential along the wrong axis (`U[0].T`) | **bit-identical** | 45.7× / 81.8× / 119.5× |

Neither repair was about dimension — a half-period wavenumber, so the mirror ghost and the periodic
ghost stop coinciding, and an asymmetry between the axes, so transposing the field is detectable. A
correct 2D test of the wall closure had been sitting there measuring nothing about the wall.

So the rule is now stated in two parts: `max|diff| = 0` in 1D means the test **must** go up a
dimension, and going up is **not sufficient**. The `d ≥ 2` bullet is scoped to match, and the closing
burden reads "in any dimension" rather than implying 1D is the only place a directional property can
hide.

Gate: **GATE GREEN**.

PR #2022, opened 2026-08-19
test: pin that `source_term` is honoured or refused, never silently discarded

### LINKED ISSUE #2020: [Design] Only one FP solver accepts `source_term`, so an MMS cannot reach the other four
The FP-side counterpart of #1991. Measured at Nx=21 on the 2D no-flux coupled MMS fixture of #2016,
each FP solver paired against `HJBFDMSolver` so a failure localises to the FP side:

| FP solver | outcome |
|---|---|
| `FPFDMSolver` | converged, 4 Picard iterations, `eu` = 1.5759e-01 |
| `FPFVMSolver` | `ValueError: operands could not be broadcast together with shapes (21,21) (882,)` — #2019 |
| `FPGFDMSolver` | `NotImplementedError: FPGFDMSolver.solve_fp_system does not accept 'source_term'` |
| `FPParticleSolver` | `NotImplementedError: ... does not accept 'source_term'` |
| `FPSLSolver` | `NotImplementedError: ... does not accept 'source_term'` |

The `NotImplementedError`s are fail-loud guards working correctly — the problem is not that they
raise, it is what the raising implies: **the manufactured-solution route reaches exactly one FP
solver.** Every convergence-order claim the suite can make about the FP side is a claim about
`FPFDMSolver`.

## Why this is not just "add the kwarg"

#1991's title says seven of ten HJB solvers drop `source_term`. Re-measured on the current tree by
`inspect.signature`, five of ten accept it — the count is stale. But the more useful correction is
that **the signature is a coarser predicate than the hazard**: `HJBWENOSolver` accepts the kwarg and
still refuses this problem, with

```
NotImplementedError: HJBWENOSolver: source_term is threaded through the 1D path only;
this problem is 2D.
```

So a census keyed on "accepts `source_term`" reports a solver as reachable when the source reaches
only its 1D branch. The predicate that matches the hazard is "a non-zero source demonstrably changes
the answer" — which is cheap to run: zeroing `s_hjb` moves `HJBGFDMSolver`'s `eu` from 1.9699e-02 to
9.7640e+00, a factor of 496, so that one genuinely threads it.

## Suggested shape

Whatever lands should come with that discrimination check per solver, not a signature audit —
otherwise the next stale table says five of ten and is wrong in the other direction.

Refs #1991 #2019 #2016

derrring:
Closes the gap #2020 names: four solvers accept `source_term` through `**kwargs` and discard it, and
nothing in the suite would notice.

## Why a signature audit cannot do this job

`inspect.signature` answers *does this callable name the parameter*, which is coarser than *does a
source reach the discretisation*. Both ways they diverge are live here:

- `HJBWENOSolver` **names** `source_term` and still refuses it in multi-D — the multi-D path is a
  dimensional split, so a source added per axis sweep would be applied $d$ times. A signature census
  counts it reachable.
- Four solvers declare `**kwargs` and swallow it. A census keyed on *does not raise `TypeError`*
  counts **those** reachable.

Keyed either way the census is wrong about one group. That is the same coarse-predicate failure that
put a row in #1991's table which was wrong the day it was written, so the predicate here is
behavioural.

## Measured

Pass a real source, watch the answer:

| solver | source calls | max&#124;with − without&#124; |
|---|---|---|
| `HJBFDMSolver` | 5 | 1.000000e+00 |
| `FPFDMSolver` | 5 | 1.000000e+00 |
| `MeshlessGalerkinHJBSolver` | **0** | **0.000000e+00** |
| `MeshlessGalerkinFPSolver` | **0** | **0.000000e+00** |
| `HJBFEMSolver` | **0** | **0.000000e+00** |
| `FPFEMSolver` | **0** | **0.000000e+00** |

The first two rows are the positive control **without which the rest prove nothing** — a harness that
never delivered a source would report every solver as swallowing it.

The last four are a **RECORDED DEFECT pin, not a specification**. They assert today's silent discard
so that fixing it — by honouring the source *or* by refusing it — trips the assertion, and the
failure message says to move the solver out of the list.

A second test classifies every solver in `mfgarchon.alg` by **MRO-resolved** signature and pins the
swallowing set, so a new solver inheriting a `**kwargs` solve method cannot join the defect quietly.
MRO and not each class's own `__dict__`: three of the six define nothing themselves and resolve to a
concrete intermediate that already dropped the parameter (`FPFEMSolver` → `WeakFormFPSolver`,
`MeshlessGalerkinFPSolver` → `WeakFormFPSolver`, `HJBFEMSolver` → `WeakFormHJBSolver`).

## Scope, surveyed rather than assumed

**The coupling layer is not the hazard.** `resolve_volatility_kwarg` treats `**kwargs` as *not*
accepting a parameter and raises; the `source_term` branch beside it has done so since #1424, and
`test_kwarg_gate_var_keyword_1783.py` pins that end-to-end on both the Picard and the Newton path.

**The hazard is the direct call**, which bypasses that gate — and no test currently exercises it.
All six direct `source_term=` call sites in the suite are on `HJBFDMSolver`, `FPFDMSolver`,
`HJBGFDMSolver`, `HJBWENOSolver` and `PenaltyHJBSolver`. The exposure is latent, so no published
convergence order is invalid. This file is what keeps that true by failure rather than by nobody
having tried.

## For whoever fixes #2020: not the base class

Both `BaseHJBSolver.solve_hjb_system` and `BaseFPSolver.solve_fp_system` **already declare**
`source_term`, and both are abstract. The contract is stated correctly. The parameter is dropped by
three concrete definers — `WeakFormFPSolver:114`, `WeakFormHJBSolver:262`,
`MeshlessGalerkinHJBSolver:82` — so six classes collapse to three owners.

Worth noting for the design: an abstract method's signature is a *declaration*, and Python does not
enforce it. A subclass may override with any signature at all, `**kwargs` included, and nothing
checks — which is the structural reason this could exist unnoticed. A base-level `__init_subclass__`
validating the override against the declared signature would make it unreintroducible, but it would
fail all six immediately, so it belongs in the same change as the fix, not here.

## Discrimination

Four mutants, each failing the expected row:

| mutant | result |
|---|---|
| reclassify a swallower as honouring | `FAILED [FPFEMSolver]` |
| reclassify a threader as swallowing | `FAILED [FPFDMSolver]` |
| drop a name from the pinned set | `FAILED test_the_swallowing_set_has_not_grown` |
| make the injected source identically zero | `FAILED [FPFDMSolver]` |

Gate: `./scripts/local_ci.sh` **GATE GREEN**. New tests 7 passed in 0.23 s.

Refs #2020 #1991 #1424 #1783

### DIFF
--- a/changelog.d/2020-source-term-channel-pin.added.md
+++ b/changelog.d/2020-source-term-channel-pin.added.md
@@ -1,0 +1,38 @@
+A behavioural pin that `source_term` is either honoured or refused, never silently discarded.
+
+`inspect.signature` answers "does this callable name the parameter", which is coarser than "does a
+source reach the discretisation", and the repo contains both ways the two diverge: `HJBWENOSolver`
+names the parameter and still refuses it in multi-D, while four solvers declare `**kwargs` and
+swallow it whole. A census keyed on either signature property misclassifies one group or the other,
+which is how #1991's table came to carry a row that was wrong when it was written.
+
+The new test passes a real source and watches the answer. Measured at the time of writing:
+
+| solver | source calls | max&#124;with − without&#124; |
+|---|---|---|
+| `HJBFDMSolver` | 5 | 1.000000e+00 |
+| `FPFDMSolver` | 5 | 1.000000e+00 |
+| `MeshlessGalerkinHJBSolver` | 0 | 0.000000e+00 |
+| `MeshlessGalerkinFPSolver` | 0 | 0.000000e+00 |
+| `HJBFEMSolver` | 0 | 0.000000e+00 |
+| `FPFEMSolver` | 0 | 0.000000e+00 |
+
+The last four rows are a **recorded defect pin, not a specification** (#2020): they assert the
+current silent-discard so that fixing it — by honouring the source or by refusing it — trips the
+assertion, and the failure message says to move the solver out of the list.
+
+A second test classifies every solver in `mfgarchon.alg` by MRO-resolved signature and pins the set
+that swallows through `**kwargs`, so a new solver inheriting a `**kwargs` solve method cannot join
+the defect silently. Classification is by MRO and not by each class's own `__dict__`: three of the
+six define nothing themselves and resolve to a concrete intermediate that already dropped the
+parameter.
+
+Scope, established by survey rather than assumed: the coupling layer is not the hazard —
+`resolve_volatility_kwarg` treats `**kwargs` as not accepting a parameter and raises, and the
+`source_term` branch beside it has done the same since #1424. The hazard is the direct
+`solve_*_system` call, and no test in the suite currently drives a manufactured solution through a
+swallowing solver.
+
+Four mutants confirm the pin discriminates: reclassifying a swallower as honouring, reclassifying a
+threader as swallowing, dropping a name from the pinned set, and making the injected source
+identically zero each fail the expected row.
--- a/tests/unit/test_alg/test_source_term_channel_2020.py
+++ b/tests/unit/test_alg/test_source_term_channel_2020.py
@@ -1,0 +1,257 @@
+"""Issue #2020: `source_term` must be honoured or refused, never silently discarded.
+
+WHY THIS IS NOT A SIGNATURE AUDIT
+---------------------------------
+`inspect.signature` answers "does this callable name the parameter", which is a coarser question
+than "does a source reach the discretisation". Two ways it diverges, both live in this repo:
+
+- `HJBWENOSolver` NAMES `source_term` and still refuses it in multi-D (`hjb_weno.py`), because the
+  multi-D path is a dimensional split and a source added per axis sweep would be applied `d` times.
+  A signature census counts it as reachable.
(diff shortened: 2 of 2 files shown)


derrring:
Folded into #2025 and closed.

This PR asserts four solvers **swallow** `source_term`; #2025 makes them **honour** it. The two are
reciprocal, so no merge order leaves both true — whichever landed second would break the other. That
is an ordering hazard rather than a disagreement, and the fix is to put the behavioural test in the
same change that makes it true.

Following this PR's own retirement condition — *"Move the solver out of the 'swallows' rows and
delete the pin if the list empties"* — all four rows are now `honours` and `_SWALLOWERS` is empty.

**The behavioural test survives in #2025 and is not redundant with the `__init_subclass__` gate
added there.** The gate reads a *signature*. This repository has twice had a solver whose signature
and behaviour disagreed: `HJBWENOSolver` names `source_term` and refuses it in multi-D, and six
solvers named nothing and swallowed it whole. A signature gate cannot separate those; passing a real
source and watching the answer can. The emptied `_SWALLOWERS` set is the behavioural witness that the
structural gate does what it claims — which a class-definition gate cannot check about itself.

Superseded by #2025.

Issue #2023, opened 2026-08-19
WeakFormHJBSolver's default (Picard) path applies the Hamiltonian with the wrong sign


derrring:
`WeakFormHJBSolver.solve_hjb_system`'s Picard branch — **the default**, `use_newton: bool = False` —
adds the Hamiltonian to the right-hand side instead of subtracting it, so the whole of `H` enters the
solve negated. The Newton branch twenty lines up has it right, as does `HJBFDMSolver`.

Affects `HJBFEMSolver` and `MeshlessGalerkinHJBSolver`, both of which reach this code.

## The mechanism

`weak_form_hjb_solver.py:347`

```python
rhs = (self._M / dt) @ U[n + 1]
...
rhs += self._M @ H_values        # <- should be -=
```

The equation is `-u_t + H - D·Lap(u) = 0`. Backward Euler, weak form:

```
M(U[n] - U[n+1])/dt + M·H + D·K·U[n] = 0
  =>  (M/dt + D·K) U[n] = (M/dt) U[n+1] - M·H
```

so `H` moves to the RHS with a sign flip. The Newton residual at `:214` is
`(M/dt)(U_cur - U_next) + D·K·U_cur + M·H - rhs_coupling`, which is the same equation written
without the move — and it has the sign right.

## Measured

Fixture with an exact answer and no discretization error in the quantity measured: `H = c` constant
(quadratic control cost with a constant coupling, `m ≡ 1`), `u_T = 0`, no-flux. Then `p ≡ 0` and
`Lap u ≡ 0`, so `u'(t) = c` and `u(0) = -c·T` exactly.

| c | analytic | **FEM picard** | FEM newton | FDM | meshless |
|---|---|---|---|---|---|
| 1.0 | −0.200000 | **+0.200000** | −0.200000 | −0.200000 | **+0.200000** |
| 2.0 | −0.400000 | **+0.400000** | −0.400000 | −0.400000 | **+0.400000** |
| −3.0 | +0.600000 | **−0.600000** | +0.600000 | +0.600000 | **−0.600000** |

Exactly `−1×` at every value, so it is a sign inversion and not an offset or a discretization
artifact. Spatial spread of the returned field is ~1e-16, i.e. the solution really is constant as the
fixture intends.

With `rhs -= self._M @ H_values` all four columns equal the analytic value at all three `c`.

## Provenance and why nothing caught it

[...]

Issue #1316, opened 2026-06-11
HJB solvers accept volatility_field but ignore it → HJB/FP diffusion mismatch for spatial volatility (Issue #1248 Phase 2)


derrring:
## Bug
The coupling iterators forward `volatility_field` to `solve_hjb_system` when the solver's signature declares it (`coupling/base_mfg.py:75-76`, signature-gated). `HJBGFDMSolver`, `HJBSemiLagrangianSolver`, `HJBWENOSolver` all **declare** `volatility_field` in their `solve_hjb_system` signature + docstring (e.g. `hjb_gfdm.py:2912/2932`) but **never use it** — the body resolves diffusion via `_get_sigma_value()` / `problem.sigma` only. (`HJBHowardSolver` swallows it via `**_unused`, `hjb_howard.py:~675`.)

Effect: when a user passes a spatially-varying `volatility_field` (array/callable) to the iterator, **FP solves with the spatial field while HJB silently solves with scalar `problem.sigma`**, breaking the Picard fixed-point correspondence. The scalar-mismatch case is warned at `fixed_point_iterator.py:177-188`, but the array/callable case slips past that guard (it only checks `isinstance(.,(int,float))`). This is the deferred consumption-side Phase 2 of #1248/#1296 (the forwarding was fixed; HJB consumption was not), and it is not currently tracked by an open issue.

### Related concrete instance (same root)
`HJBGFDMSolver._get_sigma_value` (`hjb_gfdm.py:2881-2884`): for a **callable** `problem.sigma` on the batch path (`point_idx=None`, hit by the vectorized residual at ~line 2085), the comment says "use representative value (center of domain)" but the code returns a hardcoded `1.0` — so callable σ is silently replaced by σ=1.0 on the main batch path (≈2× diffusion error for typical σ(x)).

## Fix
Either (a) implement `volatility_field` consumption in the HJB solvers (route it through `_get_sigma_value` as the diffusion source), or (b) fail loud if a non-None `volatility_field` reaches a solver that cannot honor it (don't silently accept-and-ignore). Replace the `_get_sigma_value` callable-σ `1.0` fallback with an actual center-of-domain evaluation. Add a Picard convention-agreement test (spatial σ → HJB and FP use the same diffusion).

Found in the 2026-06-11 triage sweep; verified on v0.19.8 (iterator forwards at base_mfg.py:75; gfdm declares+ignores at hjb_gfdm.py:2912). Refs #1248.

PR #2025, opened 2026-08-19
feat: thread `source_term` through the weak-form family, and gate the signature (#2020)

derrring:
**Stacked on #2024** (the Picard sign fix), because the source's sign is fixed relative to $H$'s —
placing $S$ against an inverted $H$ would have baked the inversion in. Review #2024 first.

Closes the gap #2020 names: six solvers accepted `source_term` through `**kwargs` and discarded it —
no exception, no warning, bitwise-identical output.

## Two changes cover all six

| definer | reached by |
|---|---|
| `WeakFormFPSolver.solve_fp_system` | `FPFEMSolver`, `MeshlessGalerkinFPSolver` |
| `WeakFormHJBSolver.solve_hjb_system` | `HJBFEMSolver`, and `MeshlessGalerkinHJBSolver` (forwards) |

The HJB side needed no new machinery. `_solve_timestep_newton` already had an `rhs_coupling`
parameter its residual **subtracts**, and the documented contract is
`F(u) = (u - u_next)/dt + H - S` — so that is the source slot.

## Verified by order, not by liveness

"Passing a source changes the answer" is a liveness check; what a source channel owes is an **order**
against an exact solution. Manufactured solutions on $[0,1]$ with no-flux walls, with the spatial and
temporal studies **deliberately separated** — measured together the smaller order caps the larger and
neither is visible:

| study | HJB picard | HJB newton | FP |
|---|---|---|---|
| space, P1, steady $u^\*$ | 1.883 / 1.945 / **1.974** | 1.818 / 1.805 / **1.895** | 2.021 / 2.014 / **2.008** |
| time, backward Euler | 0.999 / 1.000 / **1.000** | 1.039 / 1.018 / **1.009** | 0.948 / 0.973 / **0.987** |

$O(h^2)$ in space for P1 and $O(\Delta t)$ in time for backward Euler, on both HJB branches.

The steady/time split is not cosmetic: with a steady $u^\*$ the time term vanishes identically, so
backward Euler is exact and the residual is purely spatial.

**A trap the test file now pins.** The first temporal fixture was `a(t) = 1 + (T - t)` — linear in
$t$. Backward Euler's truncation error is $(\Delta t/2)\,u_{tt}$, identically zero there, so the
study measured the spatial floor and reported **order 0.02 from a correct implementation**.
`test_the_time_fixture_is_not_degenerate` asserts $u_{tt} \neq 0$, because a fixture that cannot
express the error is indistinguishable from a scheme that has none.

Mutation-checked: flipping the manufactured source's sign, making the time fixture linear again, and
zeroing the solver's source application each fail the expected test.

## The gate

An abstract method's signature is a **declaration** and Python does not enforce it — a subclass may
override with `(*args, **kwargs)` and nothing checks. That is the structural reason this could exist
unnoticed, and it is not fixed by threading the parameter.

`BaseHJBSolver` and `BaseFPSolver` now validate overrides at class-definition time: **an override
that accepts `**kwargs` must NAME `source_term` and `volatility_field`.**

Scoped to those two because they have the incident history (#1424, #2020; #1316, #1783). A blanket
"name every declared parameter" rule is **not satisfiable** — the base declares
`m_initial_condition` while seven implementations use `M_initial` and three use `m_initial`, so it
would be a rename, not a gate. Solvers without `**kwargs` are untouched: an unnamed parameter there
already raises `TypeError`, which is loud.

A solver that cannot support a parameter still names it and raises inside — which is what
`HJBWENOSolver` does for multi-D `source_term`. Refusal is a behaviour; an absent signature is not.

## Two #1783 tests repaired, with more coverage than before

They asserted the meshless pair is **refused** a `volatility_field`, guarded by *"this solver must be
the `**kwargs` shape, or the test does not exercise the hole"* — i.e. they depended on a production
solver keeping a hole, and the author flagged it.

Naming the parameter closes that hole, and the solver was never unable to consume the field:

| `volatility_field` | movement vs the `sigma=0.3` default |
|---|---|
| `0.3` | **0.000000e+00** — correct no-op |
| `0.9` | 4.369684e-01 |
| array 0.5→0.9 | 2.846775e-01 |

So the gate should now **forward** to this pair, and does — pinned by spying on both sides. The
refusal half moved to a **stub** that really has the `**kwargs` shape, deliberately not a
`BaseHJBSolver` subclass (the new gate refuses that shape at class definition, and the coupling layer
duck-types). That also stops this test being invalidated again by the next solver that widens its
signature.

One detail worth flagging for review: the spies use `functools.wraps`, and it is **load-bearing**.
The gate reads `inspect.signature` of what it is about to call, so a bare wrapper presents
`(*args, **kwargs)` and is itself refused — the instrument measuring itself. That cost one red gate
run to find.

Gate: `./scripts/local_ci.sh` **GATE GREEN**.

Refs #2020 #2023 #1424 #1783 #1316

### DIFF
--- a/changelog.d/2020-weak-form-source-term.added.md
+++ b/changelog.d/2020-weak-form-source-term.added.md
@@ -1,0 +1,42 @@
+`source_term` now reaches the weak-form solver family, and a class-definition-time gate stops a
+future solver from swallowing it.
+
+Six solvers accepted `source_term` through `**kwargs` and discarded it — no exception, no warning,
+bitwise-identical output. They reduce to two definers, so two changes cover all six:
+`WeakFormFPSolver.solve_fp_system` (used by `FPFEMSolver`, `MeshlessGalerkinFPSolver`) and
+`WeakFormHJBSolver.solve_hjb_system` (used by `HJBFEMSolver`, and by
+`MeshlessGalerkinHJBSolver`, which forwards to it).
+
+The HJB side needed no new machinery: `_solve_timestep_newton` already had a `rhs_coupling`
+parameter that its residual **subtracts**, and the documented contract is
+`F(u) = (u - u_next)/dt + H - S`, so that is the source slot.
+
+**Verified by order, not by liveness.** "Passing a source changes the answer" is a liveness check.
+Measured against manufactured solutions on `[0, 1]` with no-flux walls, with the spatial and
+temporal studies deliberately separated so neither caps the other:
+
+| study | HJB picard | HJB newton | FP |
+|---|---|---|---|
+| space, P1, steady `u*` | 1.883 / 1.945 / 1.974 | 1.818 / 1.805 / 1.895 | 2.021 / 2.014 / 2.008 |
+| time, backward Euler | 0.999 / 1.000 / 1.000 | 1.039 / 1.018 / 1.009 | 0.948 / 0.973 / 0.987 |
+
+$O(h^2)$ in space for P1 and $O(\Delta t)$ in time for backward Euler, on both HJB branches.
+
+The test file also pins the trap that produced those numbers on the second attempt: the first
+temporal fixture was linear in `t`, and backward Euler's truncation error is $(\Delta t/2)u_{tt}$,
+identically zero there. It reported order 0.02 from a correct implementation, because it was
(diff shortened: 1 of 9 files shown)


derrring:
Auto-closed by GitHub when its base branch was deleted on merging #2024. Continued as **#2026** — same branch, same content, rebased onto the new `main`. The review discussion stays here.

Issue #2038, opened 2026-08-21
A discrimination mutation's verify probe went blind, and a dead instrument reports exactly what a fixed defect reports


derrring:
Found while re-measuring the kill matrix (it was stamped at `f961db90`, 6138 collected; the suite is now 6384).

## The observation that looks like good news and is not

`bc_uniform_dispatch_reads_as_mixed` went from **31 killers to 0**, the only survivor in the matrix. Every other row moved the other way — 18 of 24 rows gained killers, total (mutation, test) pairs 582 → 735, killer tests 449 → 568.

A kill count is a claim about the **suite**. Read literally, this row says a convention 31 tests used to notice is now completely unguarded.

It says nothing of the kind. The row's `status` is `INEFFECTIVE` and its `seconds` is `0.0` — the suite never ran, because the mutation no longer changes behaviour the probe can see.

## Diagnosis

The mutation replaces `if bc.is_uniform:` with `if False:` in `mfgarchon/geometry/boundary/applicator_fdm.py:1035`, forcing every BC down the per-face path. Its `verify` probe used a Neumann BC.

Measured, same input under mutant and clean code:

| case | clean | mutant | differs |
|---|---|---|---|
| 1D Neumann `v=2`, depth 1 | `[1.1 1. 2. 3. 3.1]` | `[1.1 1. 2. 3. 3.1]` | no |
| 1D Neumann `v=2`, depth 2 | — | — | no |
| 1D Dirichlet `v=5` | — | — | no |
| 1D no-flux | — | — | no |
| 2D Neumann `v=2` | — | — | no |
| **1D Robin `α=1, β=1, g=3`** | `[1.097561 1. 2. 3. 3.]` | `[5. 1. 2. 3. 3.]` | **yes** |

So the two paths **have converged on Neumann** — and on Dirichlet, no-flux and 2D — while **Robin still separates them, by a wide margin**. The divergence the mutation models is alive; the Neumann witness died.

The mutation target string is still present and still applies cleanly (verified: 1 occurrence at `applicator_fdm.py:1035`). Nothing about the mutation is stale except what it looks at.

## Why this is the failure mode worth naming

`verify` exists precisely to distinguish *"nothing catches this"* from *"the mutation never took effect"* — the module docstring says so. **It is itself vulnerable to the same confusion one level down**: an instrument that has stopped being able to see produces a reading identical to the one a genuinely fixed defect produces. Zero, either way.

[...]

PR #2039, opened 2026-08-21
fix(discrimination): a verify probe went blind; make verify two-sided; re-measure the matrix (#2038)

derrring:
Three commits. The kill matrix was stamped at `f961db90` / 6138 collected; the suite is now 6384 and the gate had been printing its fraction with a `STALE` marker.

## The alarming row that was not a fact about the suite

A first sweep returned `bc_uniform_dispatch_reads_as_mixed` at **31 killers → 0**, the only survivor, while 18 of 24 rows moved the other way. A kill count is a claim about the *suite*, so read literally that row said a convention 31 tests used to notice had become unguarded.

Its `status` was `INEFFECTIVE` and its `seconds` `0.0` — the suite never ran, because the probe could no longer see the mutation take effect. Re-pointed, the row measures **43**.

## What review corrected in the first diagnosis

I claimed *"the uniform and per-face ghost paths now agree bit-for-bit on Neumann"*. **False**, and confirmed false by measurement:

| input | clean | mutant |
|---|---|---|
| Neumann, **static scalar** `2.0` | `[1.1 1. 2. 3. 3.1]` | same |
| Neumann, **callable** `lambda t: 2.0` | `[1.1 …]` | `[1. …]` |
| Robin `α=1, β=1, g=3` | `[1.097561 …]` | `[5. …]` |
| Neumann, **order 3** | `3.490909` | `1.1` |

The dead probe asserted `== 1.0`, which is exactly what the mutant returns for a **callable** value — it was one `lambda` from working.

**The mechanism was wrong too.** Nothing converged on Neumann. `_update_ghosts_mixed` finds no segment for a uniform BC and synthesizes `__default__` from `bc_type` and `bc.default_value` **only**, dropping `alpha`, `beta` and callable values. Static scalar Neumann is invisible to that because its ghost formula reads none of them. The Robin witness is the same drop from the other side: the mutant's `5.0` is bit-identical to a clean `dirichlet_bc(value=3.0)` — the `β=0` degenerate case, not Robin arithmetic, and independent of the α and β the probe passes.

## So Robin was a bad replacement, and is not used

A probe whose witness is a latent defect is one bugfix from blind. Measured against a `__default__` that forwards `alpha`/`beta`/`value`, **Robin returns to `1.097561` with the mutation still fully applied** — dead exactly as before.

`order > 2` is structural: `_update_ghosts_uniform` dispatches to `_apply_poly_extrapolation`, and the per-face path has **no order dispatch at all**. At order 3 it separates clean from mutant *and from mutant-plus-that-fix*. `PreallocatedGhostBuffer` and a `_ghost0` helper were added to the prelude, since `pad_array_with_ghosts` does not expose `order`. The duplicate `robin_bc` import the first commit added is reverted — it was already imported three lines below.

## The harness gap that allowed all of this

`_mutation_is_live` evaluated `verify` against the **mutated tree alone**. That proves the expression is true there and nothing else: a probe true under *both* trees passes and proves nothing, and the failure that actually happened — true under *neither* — was caught only by luck of direction.

`verify` is now **two-sided**: false before the mutation, true after, with the file restored and re-applied around the clean evaluation so the caller's backups stay authoritative. Positive control, all three directions:

```
real probe (false clean, true mutant)  -> True
tautology '1 == 1' (true under both)   -> False   <- the case the old check would pass
impossible '1 == 2' (true under neither)-> False   <- the case that happened
```

## Re-measurement

```
tests noticing >= 1 convention   449 -> 581
(mutation, test) pairs           582 -> 778
collected                       6138 -> 6384
rows that LOST killers               0
INEFFECTIVE rows                     0
survivors (0 killers)                0
```

Largest gains: `fp_initial_condition_written_at_final_index` 90 → 128, `ghost_spacing_ignored` 22 → 58, `diffusion_scalar_2x` 145 → 162, `neumann_low_wall_flux_sign` 23 → 37, `bc_uniform` 31 → 43.

Regenerated in **one** sweep as the script instructs; the first sweep is discarded rather than merged, since a matrix assembled from two runs is not a measurement. `test_discrimination_ratchet.py`: 59 passed.

## Known and not fixed here

The `__default__` synthesis dropping `alpha`/`beta`/callable values is a real defect, deliberately left standing — fixing it is a separate change, and the probe is now chosen so that fixing it will **not** blind the row. Left open in #2038: whether `verify` should carry several probes rather than one, and whether the other 23 probes should be audited the same way (the two-sided check now does this automatically for every one of them on the next sweep).

### DIFF
--- a/changelog.d/discrimination-rebaselined.changed.md
+++ b/changelog.d/discrimination-rebaselined.changed.md
@@ -1,0 +1,7 @@
+Re-measured the discrimination kill matrix and baseline, stale since `f961db90` (6138 collected;
+the suite is now 6384). Every one of the 24 conventions held or improved — no row lost killers.
+Tests that notice at least one convention: 449 → 581; (mutation, test) pairs: 582 → 778. Also
+fixed `bc_uniform_dispatch_reads_as_mixed`, whose `verify` probe had gone blind: the uniform and
+per-face ghost paths now agree on Neumann bit-for-bit, so the probe could not see the mutation take
+effect and the row scored INEFFECTIVE — which looks identical to "nothing catches this". Re-pointed
+at Robin, where the paths still diverge widely, the row reads 43 (#2038).
--- a/scripts/discrimination_baseline.json
+++ b/scripts/discrimination_baseline.json
@@ -1,14 +1,13 @@
 {
   "_comment": "Kill counts per mutated convention. --check-baseline fails when a count DROPS (discrimination lost) and when one RISES (record the gain in the same change). Counts, not test names: this population cannot be selected by name -- the agreement-shaped patterns give 51, 114 or 156 depending on the regex.",
   "_measured_at": {
-    "collected": 6138,
-    "commit": "f961db90",
+    "collected": 6384,
+    "commit": "03e41944",
     "excluded": "tests/unit/test_discrimination_ratchet.py",
     "markers": "not slow and not benchmark and not experimental and not optional_torch and not environment and not manual",
     "paths": [
       "tests"
-    ],
-    "_note": "markers gained `not manual` in #1909; the population is UNCHANGED -- measured 6002 collected and 388 deselected under both sets, because every `manual` test also carries `slow` -- so the kill counts recorded here remain valid without a re-sweep"
+    ]
   },
   "mutations": {
     "bc_default_reads_as_reflect": {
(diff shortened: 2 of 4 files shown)


derrring:
**Audit of the other 23 probes — done, and it cost seconds rather than a re-sweep.**

The matrix committed here was measured while `_mutation_is_live` was still one-sided, so its liveness claims deserved re-checking under the new rule: some *other* probe could have been true under both trees, counted as live, and proved nothing.

That does not need a mutation run at all. The failing direction is *true on clean*, so evaluating every `verify` against the untouched tree settles it:

```
probes evaluated on the CLEAN tree:                        24
correctly FALSE on clean (good):                           24
TRUE on clean (cannot distinguish the trees):               0
```

So every row in the committed matrix is live in the two-sided sense, and no re-sweep is required to stand behind it.

This closes follow-up 3 in #2038 — *"the other 23 probes have the same exposure … a cheap audit: for each, assert the probe's value differs between clean and mutated code"* — and it is now permanent rather than a one-off, since the two-sided `_mutation_is_live` performs exactly this check on every mutation of every future sweep.

Follow-ups 1 and 2 from #2038 remain open: whether `verify` should carry several probes rather than one, and whether the paths' agreement on *static scalar* Neumann deserves its own pin.

Issue #2019, opened 2026-08-19
`source_term_fp` is called with two incompatible argument conventions depending on the FP solver


derrring:
## The measurement

The same `source_term_fp` callable, on the same problem, receives a different argument depending
only on which FP solver is attached. Probed by instrumenting the callback and printing what arrives:

| caller | `type(x)` | `np.asarray(x).shape` |
|---|---|---|
| `source_term_hjb`, always | `ndarray` | **(441, 2)** |
| `source_term_fp` via `FPFDMSolver` | `ndarray` | **(441, 2)** |
| `source_term_fp` via `FPFVMSolver` | `tuple` | **(2, 21, 21)** |

On a `TensorProductGrid(bounds=[(0,20)]*2, Nx_points=[21,21])`, `geometry.get_spatial_grid()` is
`(441, 2)` and `geometry.meshgrid()` is a length-2 tuple of `(21,21)` arrays. `fp_fvm.py:521` uses
the second:

```python
source_grid = self.problem.geometry.meshgrid() if source_term is not None else None
...
m = m + dt * np.asarray(source_term(k * dt, source_grid), dtype=float)
```

while the FDM path (and the whole HJB side) uses the flat point list.

## Consequence

A callback written against the documented `(N, d)` convention returns an `(N,)` array. Added to the
FVM state of shape `(21,21)` that is:

```
ValueError: operands could not be broadcast together with shapes (21,21) (882,)
```

882 = 2 x 441 — the two coordinate planes flattened together, which is what a `(N,d)`-shaped
callback does when handed a `(d, *shape)` stack.

## Why it is worse than a crash

**In 1D the two conventions coincide.** `meshgrid()` returns one `(21,)` array and
`get_spatial_grid()` returns `(21, 1)`; a callback that ravels its input accepts both. So the same
callback runs clean in 1D, produces a number, and only announces the mismatch when the problem
becomes 2D. A 2x2 control, run on this repo at Nx=21:

| | 1D no source | 1D + source | 2D no source | 2D + source |
|---|---|---|---|---|
| `FPFVMSolver` | OK, 6 iter | OK, 6 iter | OK, 4 iter | **ValueError** |
| `FPFDMSolver` | OK, 5 iter | OK, 5 iter | OK, 4 iter | OK, 4 iter |

Seven of eight cells run, so this is neither a general 2D defect nor a general source defect — it is
the interaction, and the 1D row is what hides it.

[...]

PR #2043, opened 2026-08-21
test(alg): measure which solvers a source actually reaches, not which accept the kwarg (#2020)

derrring:
#2020 asks for *"that discrimination check per solver, not a signature audit — otherwise the next stale table says five of ten and is wrong in the other direction."* This is that check.

## Why a signature is the wrong predicate

`HJBWENOSolver` names `source_term` and, before #2029, still refused a 2-D problem with `source_term is threaded through the 1D path only`. A census keyed on "accepts the kwarg" calls such a solver reachable when the source reaches one branch of it.

So the predicate is behavioural and finite: **solve twice — zero source, strong source — and ask whether the answer moved.**

| verdict | n | solvers |
|---|---|---|
| `THREADS_IT` | 5 | `HJBFDM`, `HJBGFDM`, `HJBWENO`, `FPFDM`, `FPFVM` |
| `REFUSED` | 6 | `HJBSemiLagrangian`, `FPGFDM`, `FPParticle`, `FPSL`, `FPSLAdjoint`, `FPSLJacobian` |
| `ACCEPTS_AND_IGNORES` | 0 | — the class this file exists for |
| `NOT_PROBED` | 10 | FEM ×2, meshless ×2, network ×3, weak-form ×2, penalty — **named, not inferred** |

`REFUSED` is the fail-loud guard working and is not a defect. `ACCEPTS_AND_IGNORES` is the silent-wrong case: an MMS built on such a solver measures the order of an equation *without* its source and reports a clean number.

Note `FPFVM` now threads it — #2019's broadcast failure, recorded in the issue, is fixed.

## Two holes in the instrument, both found by mutating it

Neither would have been found by reading it, and the first shipped in a draft.

**The population predicate was wrong.** Walking only the `hjb_solvers` and `fp_solvers` packages found **5 and 7**. The correct root, `alg.numerical`, gives **10 and 11** — which is where #1991's "ten" comes from. The narrow predicate silently halved the population and nothing in the output said so. Now pinned by `test_the_population_is_ten_hjb_and_eleven_fp`.

**The verdict was wrong for stochastic solvers.** An `ACCEPTS_AND_IGNORES` mutant of `FPParticleSolver` — `source_term` added to the signature, never used — went **undetected**:

```
FPParticleSolver  same-input repeat, max|diff| = 1.332e-01   STOCHASTIC
FPFDMSolver       same-input repeat, max|diff| = 0.000e+00   deterministic
```

"The answer moved" scored it `THREADS_IT` with the source ignored entirely. The verdict is now read against a **per-solver noise floor** from a same-source repeat, and the same mutant is correctly classified `INDISTINGUISHABLE_FROM_ITS_OWN_NOISE`.

A first mutation attempt also *landed* — grep confirmed the marker — and still proved nothing, because inserting `source_term` as the first positional parameter rebound the caller's arguments. Landing is not validity.

## Mutation results

| mutant | fires |
|---|---|
| `FPFDMSolver` drops the source before use | 3 tests, incl. the positive control |
| `FPParticleSolver` accepts and ignores (trailing kwarg) | the ratchet, via `INDISTINGUISHABLE` |
| reverted | 5 passed |

## Scope

Measurement only — **no solver gains `source_term` here**, and #2020 stays open for the wiring. `HJBHowardSolver` is excluded on purpose: an iterator, not a solver, and already listed under `OUTSIDE_EVERY_PREDICATE` in `scripts/capability_census.py`.

Not folded into that census: it is a **declaration** census, and its own docstring records that its one behavioural lane was removed after four failed attempts to state a discrimination rule for limit behaviour, with 41% of a 32-mutation sweep surviving the ratchet built over it. This rule is finite and exact — did a byte change — which is why it sits beside that script rather than inside it.

30 passed across the reachability, wiring, fail-loud and census suites.

### DIFF
--- a/changelog.d/2020-source-term-reachability.added.md
+++ b/changelog.d/2020-source-term-reachability.added.md
@@ -1,0 +1,10 @@
+Added a **measured** account of which solvers a manufactured source can reach
+(`tests/unit/test_alg/test_source_term_reachability_2020.py`, #2020). Both prior counts were
+signature audits, and a signature is a coarser predicate than the hazard — `HJBWENOSolver` named
+`source_term` while threading it through the 1D path only. Each solver is now solved twice, with a
+zero source and a strong one, and classified by whether the answer moved: 5 thread it, 6 refuse
+loudly, 10 are out of this fixture's reach and are listed by name. The class the test exists for is
+the third one — accepting a source and returning a byte-identical answer, which would let an MMS
+measure the order of an equation without its source and report a clean, wrong number. The verdict is
+read against a per-solver noise floor from a same-source repeat, because `FPParticleSolver` differs
+by 1.3e-01 on byte-identical inputs and "the answer moved" is not evidence for a stochastic solver.
--- a/tests/unit/test_alg/test_source_term_reachability_2020.py
+++ b/tests/unit/test_alg/test_source_term_reachability_2020.py
@@ -13,8 +13,24 @@ third is the one worth having a test for:
 ``THREADS_IT``
     The answer moved. The source is genuinely applied.
 ``REFUSED``
-    ``TypeError`` or ``NotImplementedError``. The solver is out of reach, and says so. This is
-    the fail-loud guard working -- an honest refusal, not a defect.
+    The solver is out of reach and says so. Not a defect -- an honest refusal.
+
+    Two distinct mechanisms produce it, and this probe reaches only the first:
+
+    - **The signature refuses.** Every solver measured REFUSED here raises a bare
+      ``TypeError: got an unexpected keyword argument``. That is loud *by design*, not by
+      accident: ``BaseFPSolver.__init_subclass__`` says so in as many words -- *"No
+      ``**kwargs``: an unnamed parameter raises TypeError at the call site, which is loud"* --
+      and the gate exists only to stop a ``**kwargs`` signature from swallowing the argument
+      silently instead.
+    - **The coupler refuses**, with guidance. On the documented route -- a problem that defines
+      a source, passed through ``compose_fp_source`` -- ``mfg_residual.py`` raises
+      ``NotImplementedError: ... does not accept 'source_term', but the problem defines an FP
+      source term``, naming the remedy. This probe calls the solver directly, so it never gets
+      there.
+
+    Both are honest. Conflating them is not: an earlier draft of this file called the
+    ``TypeError`` "the fail-loud guard working", which named the wrong guard.
 ``ACCEPTS_AND_IGNORES``
     Took the argument, returned a byte-identical answer. **A silent wrong result**: an MMS built
     on such a solver measures the order of the wrong equation and reports a clean number.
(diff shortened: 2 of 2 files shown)


derrring:
**Blocker in this PR, found by measuring my own table rather than reading it. Do not merge as it stands.**

`HJBWENOSolver` is listed `THREADS_IT`. On a 2-D problem it still raises:

```
1D  max|diff| = 9.728e-01  -> THREADS IT
2D  RAISED NotImplementedError: HJBWENOSolver: source_term is threaded
    through the 1D path only; this problem is 2D
```

#2029 deleted that solver's duplicated 2-D/3-D *implementations*; it did not thread the source through them. The refusal quoted in #2020's own body is still live.

So the fixture is 1-D, the verdict is 1-D, and the table carries **no dimension**. That is precisely the failure #2020 says a signature audit commits — *"a census keyed on 'accepts source_term' reports a solver as reachable when the source reaches only its 1D branch"* — and a behavioural probe on a single-dimension fixture commits it just as cleanly. The predicate got better; the fixture stayed the same shape as the hazard.

**What it needs:** the classification has to run per dimension, with a verdict like `THREADS_1D_ONLY` for exactly this case, and `NOT_PROBED` recorded per dimension too — several of the ten unreachable solvers are unreachable for reasons that differ between 1-D and 2-D.

Landing the fix rather than a caveat: a note saying "measured in 1-D" would leave the table's one wrong row wrong, and this row is the issue's own worked example.

An audit of the rest of the instrument is running; the remaining findings will land in the same revision.

derrring:
**Closing this. It is a duplicate, and I should have found that before writing a line.**

`tests/unit/test_alg/test_source_term_channel_2020.py` already exists, on this same issue, with this same thesis. Its docstring:

> `inspect.signature` answers "does this callable name the parameter", which is a coarser question than "does a source reach the discretisation". … So the population predicate has to be behavioural: pass a source, and see whether the answer moves.

That is the argument of this PR, written first and written better. It also names the WENO case I reported here as a finding — **and gives the reason I did not have**:

> `HJBWENOSolver` NAMES `source_term` and still refuses it in multi-D, because the multi-D path is a dimensional split and a source added per axis sweep would be applied `d` times.

So the 2-D refusal is a deliberate correctness guard, not an unfinished branch. My blocker comment above framed it as a gap; it is a decision.

It also settles the premise this PR was built on. Its docstring records **"STATUS AFTER #2020: every solver here now HONOURS a source"**, `_SWALLOWERS` is empty, and the six that used to swallow it — `HJBFEMSolver`, `MeshlessGalerkinHJBSolver`, `WeakFormHJBSolver`, `FPFEMSolver`, `MeshlessGalerkinFPSolver`, `WeakFormFPSolver` — are **exactly six of the ten rows I recorded as `NOT_PROBED`**. They are not out of reach; my fixture could not build them. It drives all six and passes.

I built a rival census of a channel that already had one, on the issue that names it, having spent the night prosecuting single-source-of-truth defects in other people's work. The search that would have caught it is one grep over `tests/` for the issue number. I grepped for MMS fixtures instead and stopped when I found one.

## What the audit found in the meantime, since it stands regardless

A 4-lens, adversarially-verified audit surfaced 11 findings. The one that matters is not in the table but in the instrument:

**`_classify` has zero mutation coverage.** Delete the `ACCEPTS_AND_IGNORES` branch → 5 passed. Set the noise floor to `0.0` → 5 passed. The declared positive control re-implements two solves on its own `FPFDMSolver` and never calls `_classify` at all, so nothing exercised the code turning a signal into a verdict. Both mutants misclassify in the dangerous direction — a stub that names `source_term` and drops it scores `THREADS_IT` with `signal = 0.000e+00` printed in a detail string no assertion reads.

Also measured, and worth carrying to whoever extends the surviving file:

- The meshless rows are reachable. `_classify` never overrides `delta`, so it stayed at `0.1` on a cloud of spacing `0.053`; at the `delta=3.5/(n-1)` every other meshless call site in the suite uses, both meshless solvers thread a source with signal `9.7e-01`. The recorded reason blamed the node set for a defaulted keyword.
- `NetworkPolicyIterationHJBSolver` does not "require a network problem" — it raises `AttributeError: 'MFGProblem' object has no attribute 'num_nodes'`, an unguarded access.
- `FPSLSolver` and `FPSLAdjointSolver` are two rows over one method object; the alias overrides only `__init__`.
- A sign-flipped source on `FPFDMSolver` raises `ValueError` from the mass-fabrication gate, which my classifier would have filed as `NOT_PROBED` — "says nothing about the solver, only about this fixture" — when it is a source-channel result.

## Next

Fold the coverage this fixture genuinely adds — the ~11 solvers the surviving file does not parametrise — into `test_source_term_channel_2020.py`, as rows on the owner rather than a second owner. Opening that separately. This branch is deleted.

Issue #2046, opened 2026-08-21
The obstacle constraint cannot reach the timestep solve, so it clips instead of solving the VI — and only HJBFDMSolver can carry one at all


derrring:
The architectural half of #2002, separated so the retirement of `PenaltyHJBSolver` could land without it.

## Where the constraint can and cannot go today

| mechanism | reaches | can it enforce `v >= Psi`? |
|---|---|---|
| `problem.obstacle` → penalty in `compose_hjb_source` | every solver | **no** — `source_term` is `(t, x) -> array`; no `v` |
| `PenaltyHJBSolver` wrapper | any inner solver | **no** — same channel; retired in #2002 |
| `solver.constraint` → `ObstacleConstraint.project` | **`HJBFDMSolver` only** | partially — see below |

`grep -rl "self.constraint" mfgarchon/alg/numerical/{hjb,fp}_solvers/` returns exactly one file. So there is currently **no mechanism that both reaches more than one solver and can enforce the constraint.**

## Why "partially" for the one that works

`ObstacleConstraint.project` reads `u` and enforces `u >= psi` on what it returns. But it is applied outside the solve:

- **1-D**, `hjb_fdm.py:497-499`: the loop runs *after* `solve_hjb_system_backward` has returned. Measured — `U_constrained` is byte-identical to `np.maximum(U_free, psi)`, with the constraint genuinely binding at `min(U_free - psi) = -0.255176`. The backward sweep never sees the obstacle, so the free boundary is not resolved and the value away from the contact set never feels it.
- **n-D**, inside `_solve_single_timestep`: genuine feedback per step, but `U[Nt]` comes straight from `U_terminal` and is never projected — measured violation `-0.244628` on the terminal slice.

Both are #2036. This issue is the reason underneath them.

## Where a constraint actually belongs

The obstacle problem is `min(-d_t u + H(x, m, grad u) - (sigma^2/2) Lap u, u - psi) = 0`. Every standard scheme — penalty with `eps -> 0`, PSOR, projected/semi-smooth Newton — needs `u` **during** the iteration, not after it. A projection applied to a converged iterate is a clip; the same projection applied to *each Newton iterate* is a projected Newton method and actually solves the VI.

[...]

Issue #2002, opened 2026-08-18
obstacle: documented as a variational inequality, implemented twice as a v-free penalty, the two 1e10 apart


derrring:
`obstacle` is documented as a variational inequality and implemented, twice, as a penalty that cannot enforce one. The two implementations differ by a factor of 1e10, while a comment states they match.

## The constraint is not enforceable by either expression

`mfg_problem.py:569` declares the intent:

```
# obstacle: Callable(x) -> array for variational inequality v >= Psi(x)
```

To penalise violation of `v >= Psi` you need `max(0, Psi - v)`: positive exactly when the constraint is broken, zero otherwise. Both implementations compute `max(0, Psi(x))`, which contains no `v` at all — it is positive wherever `Psi > 0` regardless of whether the constraint holds. **It penalises position, not violation.**

| path | expression | default scaling | ψ = 0.5 gives |
|---|---|---|---|
| `source_composition.py:130-133` | `(1.0 / eps) * max(0, psi)` | `eps = 1e6` | 5.0e-07 |
| `hjb_penalty.py:144-145` | `penalty_param * max(0, psi)` | `penalty_parameter = 1e4` | 5000.0 |

Same obstacle, same ψ, **1e10 apart** — and one divides by its parameter while the other multiplies.

## Three claims in one comment, two of them false

`source_composition.py:29-32`:

> **Obstacle** uses the approximate `v = 0` penalty `(1/eps) * max(0, psi)` (`eps = problem._penalty_eps` if set, else `1e6`). Proper handling is the `PenaltyHJBSolver` wrapper (#924); both coupling paths use this same approximation so they **match** rather than silently diverge.

1. "uses the approximate `v = 0` penalty" — **true**, and creditably explicit.
2. "Proper handling is the `PenaltyHJBSolver` wrapper" — **false**. That wrapper carries the same stub, and says so in its own comment: *"For now, we apply a static obstacle penalty."* The pointer sends a reader to a second copy of the defect.
3. "both coupling paths use this same approximation so they **match** rather than silently diverge" — **false** in the sense that matters. Same formula shape, scalings 1e10 apart. They diverge numerically, and the sentence asserting they do not is what would stop someone checking.

`hjb_penalty.py`'s own comment is worth reading in full because it refutes itself mid-paragraph: it first says the inner solver's time-stepping handles the `v` dependence, then — *"However: the source_term signature is (t, x) -> array, not (t, x, v) -> array"* — concludes it does not.

[...]

Issue #1642, opened 2026-07-19
Model declares the variational surface (coupling_cost F, terminal_coupling G) and nothing reads it — silently ignored since #875


derrring:
## Summary

`Model` — the v1.0 primary API — declares two variational fields and **no code in the library reads either**. A user can supply them, the `Model` constructs, and they are silently discarded.

```python
# mfgarchon/core/model.py:52-53
coupling_cost: Callable | None = None       # F(m) -- interaction cost (variational formulation)
terminal_coupling: Callable | None = None   # G(x, m(T)) -- terminal cost in objective (variational)
```

Reader census across `mfgarchon/`:

| field | readers |
|---|---|
| `model.hamiltonian` | 1 |
| `model.sigma` | 1 |
| `model.drift_field` | 2 |
| **`model.coupling_cost`** | **0** |
| **`model.terminal_coupling`** | **0** |

Also 0 uses in `tests/` and `examples/`, and no dynamic access (`getattr` / `asdict` / `vars` over the model all return nothing). The 4 apparent `coupling_cost` hits in `alg/reinforcement/multi_population/base_environment.py:377-411` are an unrelated local variable of the same name.

Verified: `Model(hamiltonian=..., sigma=..., coupling_cost=..., terminal_coupling=...)` constructs successfully with both fields set.

`Conditions.u_terminal` compounds it — its docstring reads *"Terminal cost u_T(x). None for variational (u is derived)"*, advertising a variational mode in which the terminal condition has different semantics. No such mode exists.

## Why this is the RFC #1574 class, one layer above everything else in that queue

This is *advertise-more-than-you-honor*: a declared capability with silence, not a loud failure, in the gap. Every other instance in the #1574 queue sits inside a solver. **This one is on `Model`, the primary user-facing API object** — the first thing a user of the v1.0 API touches.

Direct precedent: **#1285** ("MFGResidual/Newton silently ignores source_term_hjb/fp, nonlocal_operator, obstacle", CLOSED COMPLETED 2026-06-15) is the same defect on the same class of declared-but-unread problem fields. That sweep covered the Layer-1 fields and did not reach the variational pair on `Model`.

## Provenance

Introduced by `1925a388` (2026-03-29, "feat: API v1.0 Phase 2 — Model, Conditions, dual-mode MFGProblem", #875). That commit added the declarations; it wired neither. The fields have been silently ignored for the whole life of the v1.0 API.

[...]

Issue #591, opened 2026-01-17
Phase 2: Tier 2 BCs - Variational Constraints (Obstacle Problems)


derrring:
# Phase 2: Tier 2 BCs - Variational Constraints

**Part of**: #589 (Geometry & BC Architecture - Master Tracking)
**Duration**: 3-4 weeks (15-20 person-days)
**Prerequisites**: #590 (Phase 1 - Geometry Traits)

---

## Objective

Implement obstacle problems and variational inequalities (Tier 2 BCs) for MFG systems with capacity constraints, congestion modeling, and other inequality-constrained physics.

---

### 2.1 Constraint Protocol (1 week)

**Files to Create**:
- `mfg_pde/geometry/boundary/constraints.py` - Constraint base classes
- `mfg_pde/geometry/boundary/constraint_protocol.py` - Protocol definitions

**Key Abstractions**:
```python
class ConstraintProtocol(Protocol):
    def project(self, u: NDArray) -> NDArray: ...
    def is_feasible(self, u: NDArray, tol=1e-10) -> bool: ...
    def get_active_set(self, u: NDArray, tol=1e-10) -> NDArray[np.bool_]: ...

class ObstacleConstraint:
    """Unilateral constraint: u >= ψ or u <= ψ."""
    def __init__(self, obstacle, constraint_type='lower', region=None): ...

class BilateralConstraint:
    """Bilateral constraint: ψ_lower <= u <= ψ_upper."""
    def __init__(self, lower_obstacle, upper_obstacle, region=None): ...
```

**Testing**:
- Projection operator tests (idempotency, contraction)
- Active set detection accuracy

---

### 2.2 VI Solver Integration (2 weeks)

**Solvers to Modify**:
1. HJB FDM Solver (`hjb_fdm.py`)
2. HJB Semi-Lagrangian (`hjb_sl.py`)
3. Poisson Solver (for testing)

**Implementation Pattern** (Penalty Method):
```python
def solve_hjb_system(
    self,
    ...,
    constraints: list[ConstraintProtocol] | None = None,
) -> NDArray:
    for iteration in range(max_iterations):
        # Compute unconstrained update
        U_unconstrained = self._compute_hjb_update(U_prev, ...)

        # Apply constraints if present
        if constraints:
            U_new = U_unconstrained.copy()
            for constraint in constraints:
                U_new = constraint.project(U_new)
        else:
            U_new = U_unconstrained

        # Check convergence
[...]

PR #2048, opened 2026-08-21
feat(core)!: obstacle resolves as a soft wall and becomes a potential (#2002)

derrring:
Item 4 decided: **soft wall.** #2046 deferred; the variational inequality keeps a reserved slot rather than being deleted.

## Why a potential and not a source

The term is `α`-free and `u`-free — it depends on where you are, not on what you do or what the value function says. That is the definition of a potential, and it is exactly what disqualified it as the variational inequality it was documented as. So it moves to where potentials live, rather than staying in a channel that cannot carry `v`.

| | before | after |
|---|---|---|
| field | `problem.obstacle` (raises now) | `problem.state_penalty` + `state_penalty_scale` |
| sign | ambiguous — `max(0, Ψ)` with `1/eps` | **cost-signed**: positive where expensive |
| where | `compose_hjb_source` → `source_term` | Hamiltonian's `V`, at problem construction |
| VI | claimed, never delivered | reserved: `HJBFDMSolver(constraint=…)` (#591) |

`obstacle=` raises and names **both** successors, because it was conflating two different things. It also disambiguates from `obstacles` (plural) — geometric regions excluded from the domain, a different field entirely that shared almost the same name in the same class.

## The sign was measured, not derived

`potential` is **reward-signed** (G-001), so the composition **subtracts**. On a Gaussian wall at `x = 0.5`, no coupling:

| `V` amplitude | `u(0, mid)` | |
|---|---|---|
| `0` | `+0.000000` | control: flat |
| `+5` | `−1.419462` | a **positive** potential makes the wall **cheaper** |
| `−5` | `+0.555313` | a cost needs a **negative** potential |

This repo has had that sign wrong twice (#1642 B1, #1645 B2). The subtraction is therefore written **once**, in one place, and pinned by a test that reddens on inversion — verified, three of six fire under a `-` → `+` mutant.

End to end with the new field: wall on gives `u(mid) = +0.555` against a flat `0.0`, `scale=3` gives `+1.018`, and the edge moves `1.13e-03` against the middle's `5.55e-01`.

## The rival is deleted in the same change

`compose_hjb_source`'s obstacle branch is gone, along with the module docstring's account of it. That module exists to stop parallel physics paths keeping private copies of a convention; a composition that only **adds** an owner has consolidated nothing.

## Tests changed rather than adapted, where they encoded the old design

- `test_composed_hjb_source_is_byte_identical_…` — **deleted by its own retirement condition**, the last of three pins in that file to go.
- `test_issue1361_source_composition` — its oracle drops the obstacle branch (it reimplements production by hand, which is *why* it reddened), and its four obstacle cells go: that file is about **source** composition, and a soft wall is no longer a source, so a cell there would assert that a term absent from both sides is equal on both sides.
- `test_newton_solver_solves_with_obstacle` → `…_with_a_state_penalty`, and gets **stronger**. The old version checked that two couplers each remembered to consume the same source closure — a wiring fact. The new one checks that a term *inside* `H` is seen identically by a residual assembled from `H` (Newton) and by a sweep that calls `H` (Picard). The `1e-8` tolerance is unchanged against a signal that went from `5e-07` to O(1), which makes it a larger claim, not a looser one.

## One contract detail encoded rather than left to callers

`SeparableHamiltonian` evaluates the potential **per point** inside `float()`, so in 1-D `x` arrives as `array([0.3])` and a shape-`(1,)` answer raises under NumPy 2. The composition squeezes — which is what every working potential in this repo already does by hand, and now nobody has to.

50 passed across the state-penalty, composition, wiring, Newton-parity, obstacle-integration, channel and penalty suites.

**BREAKING**: `MFGProblem(obstacle=…)` raises. Callers were getting a position penalty documented as a constraint.

### DIFF
--- a/changelog.d/2002-obstacle-becomes-a-state-penalty.changed.md
+++ b/changelog.d/2002-obstacle-becomes-a-state-penalty.changed.md
@@ -10,3 +10,10 @@ while `+5` lowers it to `-1.419`. The obstacle branch is deleted from `compose_h
 same change. The **variational inequality** remains a reserved, deliberately unfinished slot:
 `HJBFDMSolver(constraint=ObstacleConstraint(...))` (#591, #2036, #2046). Note `obstacles` (plural,
 geometric regions excluded from the domain) is a different field and is unaffected.
+
+The composition COPIES the Hamiltonian rather than rebuilding it (an earlier draft rebuilt from
+five of its six constructor parameters and silently dropped `population_index`, which a
+multi-population problem needs), returns exactly what the base potential returns (so a vectorised
+base stays vectorised), and updates `_lagrangian_class` as well — `MFGComponents` snapshots the
+potential there at construction, and `HJBSemiLagrangianSolver` reads that copy whenever the control
+cost is non-smooth, so composing into the Hamiltonian alone left a solver silently unwalled.
--- a/changelog.d/2002-obstacle-becomes-a-state-penalty.changed.md
+++ b/changelog.d/2002-obstacle-becomes-a-state-penalty.changed.md
@@ -1,0 +1,12 @@
+**BREAKING.** `problem.obstacle` is retired and raises (#2002). It was documented as the
+variational inequality `v >= Psi(x)` and implemented as `max(0, Psi(x))` — a term with no `v`,
+byte-identical at a node satisfying the constraint and one violating it. It resolved as a **soft
+wall**, which is what it always computed: a cost that is `alpha`-free and `u`-free is a POTENTIAL,
+so it becomes `problem.state_penalty` (cost-signed, positive where expensive, with
+`state_penalty_scale`) and is composed into the Hamiltonian's `V` at problem construction rather
+than into `source_term`. The composition SUBTRACTS, because `potential` is reward-signed
+(G-001) — measured, not assumed: a potential amplitude of `-5` raises `u(0, mid)` to `+0.555`
+while `+5` lowers it to `-1.419`. The obstacle branch is deleted from `compose_hjb_source` in the
+same change. The **variational inequality** remains a reserved, deliberately unfinished slot:
+`HJBFDMSolver(constraint=ObstacleConstraint(...))` (#591, #2036, #2046). Note `obstacles` (plural,
+geometric regions excluded from the domain) is a different field and is unaffected.
--- a/mfgarchon/alg/numerical/coupling/mfg_residual.py
+++ b/mfgarchon/alg/numerical/coupling/mfg_residual.py
@@ -53,9 +53,9 @@ class MFGResidual:
     The residual measures deviation from this fixed point:
         F(U, M) = [HJB_solve(M) - U, FP_solve(U) - M]
 
-    Source / nonlocal / obstacle terms (Issue #1361):
+    Source / nonlocal terms (Issue #1361):
         ``source_term_hjb``, ``source_term_fp``, ``nonlocal_operator``, and
-        ``obstacle`` are composed via the single-source helpers
+        are composed via the single-source helpers
         :func:`source_composition.compose_hjb_source` /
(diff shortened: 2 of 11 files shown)


derrring:
**Review found four blockers. Three reproduced under my own measurement and are fixed; one did not, and I am reporting that rather than fixing what I cannot see.**

## Fixed

**1. The composition rebuilt the Hamiltonian instead of copying it.** It constructed a fresh `SeparableHamiltonian` from five of its six constructor parameters. The sixth is `population_index`, and it is live — a multi-population problem keyed to population *k* silently became population 0. Measured: `2 → 0`. Now `copy.copy` + set `_potential`, which preserves the type and every attribute and leaves the caller's own object unmutated (both pinned).

**2. The closure squeezed its result unconditionally**, so a vectorised base potential stopped looking vectorised to `SeparableHamiltonian`'s batch probe. Measured: `_potential_is_vectorized` `True → False`. It now reshapes the wall to the base's shape and returns `base - wall` untouched, so the base's contract survives. Pinned by a test asserting the composed closure returns exactly the base's shape for both a per-point and a batch call.

**3. The Lagrangian was left stale.** `MFGComponents.__post_init__` snapshots the potential into `_lagrangian_class` **before** the composition runs, and `HJBSemiLagrangianSolver` reads that copy whenever the control cost is non-smooth. Composing into the Hamiltonian alone therefore left a solver silently unwalled — the #2002 defect itself, in a second channel, and the constructor's `raise` could not see it because the Hamiltonian *was* a `SeparableHamiltonian` all along.

**4. `mfg_residual.py` carried the fourth and fifth copies of the claim this branch deleted** — a docstring still asserting "obstacle handling matches the Picard path exactly: the same `(1/eps) * max(0, psi)` from the one `compose_hjb_source`", and a user-facing `NotImplementedError` telling callers to "remove source_term_hjb / nonlocal_operator / **obstacle**". That file was corrected for this same sentence two PRs ago, which is rather the point.

Plus: my own comment named `_state_penalty_eps`, an attribute that does not exist anywhere (the field is `state_penalty_scale`), and both it and `state_penalty` are read **once**, at construction — now stated, because assigning either afterwards does nothing.

## Not reproduced

The claim was that a zero wall corrupts a **pre-existing** potential, freezing it at `x_batch[0]` and moving `u` by `1.91` on the semi-Lagrangian path. Two fixtures — a per-point base and a vectorised base — give correct values before and after in both, and a zero wall is a no-op on the batch path:

```
VEC base BEFORE: [ 0.  1.  0. -1. -0.]   vec=True
VEC base AFTER : [ 0.  1.  0. -1. -0.]   vec=True     (after the fix; was False)
TRUTH          : [ 0.  1.  0. -1. -0.]
zero wall is a no-op on the batch path? True
```

What I *did* measure in that area was the vectorisation flag, fixed above. The review's own report records a first run against a stale copy of the package; I cannot tell from here whether that is the difference, so this is **not-reproduced**, not refuted. Fix 3 may also account for the SL number, since that is the channel it came from.

## And one the gate caught, inside the fix

My Lagrangian sync used `hasattr(lagrangian, "_potential")` — which **silently skips** when the attribute is absent. That is the defect this whole series is about, reintroduced inside its own fix. The fail-fast baseline flagged it as a count going `107 → 108`; the baseline was left alone rather than regenerated, since it went **up** — a violation added, not violations fixed. Now `isinstance(..., SeparableLagrangian)`, with anything else raising and naming both the consequence and the remedy.

## Mutation status, stated honestly

Three of four new tests are mutation-verified: rebuild-instead-of-copy, skip-the-Lagrangian, and the shape contract each redden their own test. The fourth mutant — squeezing regardless of the base — is **equivalent** once the reshape is in place, so no test distinguishes it and none is claimed.

34 passed across the state-penalty, Newton-parity, composition, wiring and multi-population suites; full gate green on push.

